PR #53, opened 2023-11-08
fix: Replay as is stops working when previous message was based on GPT model + any addon

denys-kolomiitsev:
#353


denys-kolomiitsev:
/deploy-review
>Deployment status: success

PR #55, opened 2023-11-09
chore: Enchancement. Upgrade outdated packages

IlyaBondar:
### Description
upgrade outdated packages:
image


mikitabut:
If you want to upgrade this packages please collect changes in them, because some of them having minor patches which adding new features and maybe we can to use them

PR #63, opened 2023-11-09
fix: fix no scroll in playback textbox for long message, blinking scroll down icon and auto-scroll for long list responses

IlyaBondar:
### Description
- Issue #361: [Playback] There is no scroll in user-message box in some specific case (need to investigate, example is attached)
- Issue #354: Scroll down icon appears when answer is being generated. Scroll down appears for several milliseconds before auto-scroll is applied
- Issue #355: Auto-scroll doesn't work in chats based on GPT models when response contains a list

### DIFF
--- a/src/components/Chat/Chat.tsx
+++ b/src/components/Chat/Chat.tsx
@@ -189,7 +189,8 @@ export const Chat = memo(() => {
     (force = false) => {
       if (autoScrollEnabled || force) {
         setAutoScroll();
-        messagesEndRef.current?.scrollIntoView({
+        chatContainerRef.current?.scrollTo({
+          top: chatContainerRef.current.scrollHeight,
           behavior: 'smooth',
         });
       }
--- a/src/components/Chat/Chat.tsx
+++ b/src/components/Chat/Chat.tsx
@@ -182,6 +182,8 @@ export const Chat = memo(() => {
         messagesEndRef.current?.scrollIntoView({
           behavior: 'smooth',
         });
+        setAutoScrollEnabled(true);
+        setShowScrollDownButton(false);
       }
     },
     [autoScrollEnabled],
--- a/src/components/Chat/Chat.tsx
+++ b/src/components/Chat/Chat.tsx
@@ -45,6 +45,8 @@ import { NotAllowedModel } from './NotAllowedModel';
 import { PlaybackControls } from './PlaybackControls';
 import { PlaybackEmptyInfo } from './PlaybackEmptyInfo';
 
+const scrollThrottlingTimeout = 250;
+
 const clearStateForMessages = (messages: Message[]): Message[] => {
   return messages.map((message) => ({
     ...message,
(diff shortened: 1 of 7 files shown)


IlyaBondar:
/deploy-review
>Deployment status: success

IlyaBondar:
/deploy-review
>Deployment status: failed

IlyaBondar:
/deploy-review
>Deployment status: success

denys-kolomiitsev (dismissed):

src/components/Chat/Chat.tsx
```diff
@@ -170,44 +179,38 @@ export const Chat = memo(() => {
...
-      top: chatContainerRef.current.scrollHeight,
-      behavior: 'smooth',
-    });
-    textareaRef.current?.focus();
-  }, [autoScrollEnabled]);
+  const scrollDown = useCallback(
+    (force = false) => {
+      if (autoScrollEnabled || force) {
+        setAutoScroll();
+        messagesEndRef.current?.scrollIntoView({
```
mikitabut: Please check how this is working in overlay, because when we previously use this we had an issue that parent window also scrolled , but scrollTo is working fine with that

src/components/Chat/Chat.tsx
```diff
@@ -170,44 +179,39 @@ export const Chat = memo(() => {
     [dispatch],
   );
 
-  useEffect(() => {
-    if (!autoScrollEnabled || !chatContainerRef.current) {
-      return;
-    }
+  const setAutoScroll = () => {
```
mikitabut, line 182: useCallback?

src/components/Chat/Chat.tsx
IlyaBondar: Thank you. Yes, issue was reproducible in overlay mode. Fixed and pushed

src/components/Chat/Chat.tsx
IlyaBondar: @mikitabut , `setAutoScroll` isn't used in dependencies and doesn't use any changed or state properties(`disableAutoScrollTimeoutRef` always the same). That's why it isn't suggested to be covered into `useCallback` to memorize it (with zero dependencies)

mikitabut: Sorry, but useCallback also used to memorize function declaration and reuse it between renders. In such case every render will create new function and set it. Are you sure that we don't need to use it here?

IlyaBondar:
/deploy-review
>Deployment status: success

src/components/Chat/Chat.tsx
mikitabut: Got it
My undersatnding of useCallback was not right

mikitabut (approved):

src/components/Chat/Chat.tsx
IlyaBondar: With `useCallback` we memorize one instance of function, but function will be created each time plus array of dependencies as arguments for `useCallback` and then `useCallback` works as Singleton with dependencies i.e. we don't have any optimization

PR #66, opened 2023-11-10
feat: use storage agnostic approach for data

mikitabut:


denys-kolomiitsev (dismissed):

denys-kolomiitsev (dismissed):

src/utils/app/data/storage.ts
```diff
@@ -0,0 +1,24 @@
...
+import { errorsMessages } from '@/src/constants/errors';
+
+export const isLocalStorageEnabled = () => {
+  const testData = 'test';
+  try {
+    localStorage.setItem(testData, testData);
+    localStorage.removeItem(testData);
+    return true;
+  } catch (e: any) {
+    if (e.name === 'QuotaExceededError') {
```
IlyaBondar: here and in several other places we can check this specific `DOMException` from localStorage and don't use `any`
```suggestion
  } catch (e) {
    if (e instanceof DOMException && e.name === 'QuotaExceededError') {
```

src/store/ui/ui.reducers.ts
```diff
@@ -16,8 +16,8 @@ export interface UIState {
 
 const initialState: UIState = {
   theme: 'dark',
-  showChatbar: true,
-  showPromptbar: true,
+  showChatbar: false,
+  showPromptbar: false,
```
IlyaBondar, line 20: Do we really want to hide these panels by default?

IlyaBondar (dismissed):

src/store/ui/ui.reducers.ts
mikitabut: I think it's better to hide, because without them opened it looks ok on mobile and desktop, but when they both opened by default it looks very bad on mobile and ok on desktop  :)
image

denys-kolomiitsev (approved):

src/utils/app/data/storage.ts
IlyaBondar: The same is in browser-storage.ts and maybe somewhere else

akryvasheyeu (approved):

PR #67, opened 2023-11-10
feat: update rate endpoint according to new API

mikitabut:


denys-kolomiitsev (dismissed):

src/pages/api/rate.ts
```diff
@@ -26,11 +26,11 @@ const handler = async (req: NextApiRequest, res: NextApiResponse) => {
   try {
     const { responseId, modelId, value, id } = req.body as RateBody;
 
-    if (!id || !validate(id) || !responseId) {
+    if (!id || !validate(id) || !responseId || !modelId) {
       return res.status(400).send(errorsMessages[400]);
     }
 
-    const url = `${OPENAI_API_HOST}/v1/rate`;
+    const url = `${OPENAI_API_HOST}/v1/${modelId}/rate`;
```
astsiapanay, line 33: We should remove modelId from the request body as it's already in the path.

astsiapanay (approved):

PR #69, opened 2023-11-10
fix: replay as is stops working when previous message was based on gpt model and any addon

denys-kolomiitsev:
- Issue 353: [Replay as is] stops working if current message is based on GPT-4-0613 or GPT-3.5 when previous message was based on GPT model + any addon


denys-kolomiitsev:
/deploy-review
>Deployment status: failed

mikitabut (dismissed):

denys-kolomiitsev:
/deploy-review

>Deployment status: success

IlyaBondar (commented):
Everything looks well

src/store/models/models.reducers.ts
```diff
@@ -103,21 +103,34 @@ const rootSelector = (state: RootState): ModelsState => state.models;
...
 });
+
 const selectModels = createSelector([rootSelector], (state) => {
   return state.models;
 });
+
+const selectModel = createSelector(
+  [selectModels, (_state, modelId: string) => modelId],
+  (models, modelId) => {
+    return models.find(({ id }) => id === modelId) as OpenAIEntityModel;
```
mikitabut: you can use modelsMap instead
Moreover it's possible that model will not be found, but because of your type assertion it will be not catched

mikitabut (approved):

src/store/models/models.reducers.ts
denys-kolomiitsev: Implemented selector with modelsMap.
Additional check added

PR #176, opened 2023-11-16
fix: fix Playback empty info block style

IlyaBondar:
fixed:
- fixed width equals to message box for small titles
- smaller gap inside block
image


mikitabut (approved):

mikitabut:
/deploy-review

>Deployment status: failed

PR #194, opened 2023-11-21
e2e: added tests for prompts nested folders

irinakartun:
added tests for prompts nested folders

### DIFF
--- a/e2e/src/tests/promptExportImport.test.ts
+++ b/e2e/src/tests/promptExportImport.test.ts
@@ -10,15 +10,19 @@ import {
 } from '@/e2e/src/testData';
 import { ImportPrompt } from '@/e2e/src/testData/conversationHistory/importPrompt';
 import { UploadDownloadData } from '@/e2e/src/ui/pages';
-import { FileUtil } from '@/e2e/src/utils';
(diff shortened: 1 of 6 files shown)


irinakartun:
/deploy-review
>Deployment status: success

irinakartun:
/deploy-review
>Deployment status: success

e2e/src/tests/promptExportImport.test.ts
```diff
@@ -430,12 +453,303 @@ test('Import file from 1.4 version to prompts and continue working with it', asy
...
+      await folderPrompts
+        .getFolderPrompt(
+          nestedFolders[levelsCount].name,
+          nestedPrompts[levelsCount].name,
+        )
+        .waitFor();
+
+      await folderPrompts
+        .getFolderByName(nestedFolders[levelsCount - 1].name)
+        .waitFor();
```
IlyaBondar, line 562: @irinakartun , why do you use `waitFor()` here and in other places if you already have `await`? does it convert to async-method?

IlyaBondar (approved):

e2e/src/tests/promptExportImport.test.ts
irinakartun: it simply waits for element visibility on the page

Issue #80, opened 2023-11-13
Implement attachment functionality


VolhaBazhkova:
See https://github.com/epam/ai-dial-core/issues/29

We should have availability to add attachments to user message

The attachment should be loaded before sending. Attachment size should be up to 1 gb.

When clicking on icon modal will be opened with all attached files before and with folder structure.

Open Questions:

1. Do we need to have any UI where we select already existing attachments? - YES
2. Should we be able to select multiple attachments? - YES
3. What type of attachments should be supported? - Backend will provide type of attachments supported by model
4. Should we be able to see already loaded attachments and manage them? - YES

Design will be provided (in progress)

mikitabut:
Currently next things have been implemented:

1. For some applications we can get from listing next params `input_attachment_types` and `max_input_attachments` (For more details see https://github.com/epam/ai-dial-core/issues/34) which saying us do we need to show attach button or not
2. Also we put this attachment button under feature flag
3. When clicking on attach button user will see 2 options: 'Attach uploaded files' and 'Upload from device'
4. When clicking on 'Attach uploaded files' user will see file select modal will be able to manage already uploaded files, create folders and upload more files from there(to upload files user will see same steps as in point 5, except that after clicking attach button user will be navigated back to file select modal). 
5. When clicking on 'Upload from device' the user will be able to select files from device and put them to existing folders(or create new folders). After clicking on attach button the modal will be closed and he will see progress of file uploading

What haven't been done or done differently:

1. UI doesn't have manage files modal which opened from left side bar (Scope https://github.com/epam/ai-dial-chat/issues/211)
2. [TODO] UI doestn't allow to rename files, move files, rename folders, move folder, remove folder(remove folder doing automatically after removing all files from it and reloading the page).
3. From UI it's not possible to upload file to folder directly (Scope https://github.com/epam/ai-dial-chat/issues/211)
4. Some actions like download file, add new folder, delete file are under context menu
5. [TODO] Currently UI not supporting bulk updates for folders and files like download or delete
6. UI don't show recent updates section
7. UI don't show loader when listing for files and folders from root folders are loading (Scope https://github.com/epam/ai-dial-chat/issues/211)
8. It's not possible to edit already attached files in user message (Scope https://github.com/epam/ai-dial-chat/issues/211)

Some known issues and notes:
1. When cancelling upload folder it's not cancelling for fetch request, just on UI
2. When searching for folders and files - folders not filtered, because the list of folders is not fetched one time but fetching when you opening them

PR #195, opened 2023-11-21
feat: implement attaching files to conversations

mikitabut:
Implements #80


mikitabut:
/deploy-review

>Deployment status: success

src/components/Folder/Folder.tsx
```diff
@@ -273,23 +322,38 @@ const Folder = <T extends Conversation | Prompt>({
...
+      ? 'text-violet'
+      : 'text-blue-500',
+  );
+  const bgColor = classNames(
+    highlightColor === 'green'
+      ? 'bg-green/15'
+      : highlightColor === 'violet'
+      ? 'bg-violet/15'
+      : 'bg-blue-500/20',
+  );
```
IlyaBondar: Several ternary conditionals look ugly and it's not readable and bad practice.
You can create utility-method like this(in util-file):
```javascript
const getHighlightColor = (
        highlightColor:string, 
        greenColor:string, 
        violetColor:string, 
        defaultColor?:string
      ) => {
        switch(highlightColor) {
          case 'green': return greenColor;
          case 'violet': return violetColor;
          default: return defaultColor;
        }
      }
```
and use here and in other places
```suggestion
  const hoverIconColor = getHighlightColor(highlightColor, 'hover:text-green', 'hover:text-violet', 'hover:text-blue-500');
  const textColor = getHighlightColor(highlightColor, 'text-green', 'text-violet', 'text-blue-500');
  const bgColor = getHighlightColor(highlightColor, 'bg-green/15', 'bg-violet/15', 'bg-blue-500/20');
```
also you don't need to use `classNames` here

src/components/Chat/ChatInput/ChatInput.tsx
```diff
@@ -351,10 +375,16 @@ export const ChatInput = ({
...
       <div className="mx-2 mb-2 flex flex-row gap-3 md:mx-4 md:mb-0  md:last:mb-6 lg:mx-auto lg:max-w-3xl">
-        <div className="relative flex w-full grow flex-col" data-qa="message">
+        <div
+          className="relative m-0 flex max-h-[400px] min-h-[40px] w-full grow flex-col rounded bg-gray-100 focus-within:border-blue-500 dark:bg-gray-700"
+          data-qa="message"
+        >
           <textarea
             ref={textareaRef}
-            className="m-0 max-h-[400px] min-h-[40px] w-full resize-none rounded border border-transparent bg-gray-100 py-3 pl-4 pr-10 outline-none placeholder:text-gray-500 focus-visible:border-blue-500 dark:bg-gray-700"
+            className={classNames(
```
IlyaBondar: you don't need to use `classNames` here. Just add `pl-12` in the same string

src/components/Chat/ChatInput/ChatInputAttachments.tsx
```diff
@@ -0,0 +1,86 @@
...
+              size={18}
+            />
+          )}
+
+          <div className="flex grow justify-between gap-3 overflow-hidden">
+            <div className="flex grow flex-col overflow-hidden">
+              <span
+                className={classNames(
+                  'block max-w-full truncate',
+                  file.status === 'FAILED' && 'text-red-800 dark:text-red-400',
```
IlyaBondar: see recomended usage format [here](https://www.npmjs.com/package/classnames#usage)
```suggestion
                  { 'text-red-800 dark:text-red-400' : file.status === 'FAILED' },
```

src/components/Files/FileSelect.tsx
```diff
@@ -0,0 +1,363 @@
...
+                      placeholder={
+                        t('Search model, assistant or application') || ''
+                      }
+                      type="text"
+                      onChange={handleSearch}
+                      className="m-0 w-full rounded border border-gray-400 bg-transparent px-3 py-2 outline-none placeholder:text-gray-500 focus-visible:border-blue-500 dark:border-gray-600 dark:focus-visible:border-blue-500"
+                    ></input>
+                    <div className="flex min-h-[350px] flex-col overflow-auto">
+                      <button
+                        className={classNames(
```
IlyaBondar: You don't need to use `classNames` here - just string

src/components/Files/FileItemContextMenu.tsx
```diff
@@ -0,0 +1,73 @@
...
+    <>
+      <Menu
+        type="contextMenu"
+        trigger={
+          <DotsIcon
+            className={classNames('rotate-90 text-gray-500')}
+            width={18}
+            height={18}
+            size={18}
+          />
```
IlyaBondar: 1) You don't need to use `classNames` here
2) why we still use `DotsIcon` with rotation instead of `IconDots` from `@tabler/icons-react`? You won't find difference
```suggestion
          <IconDots
            className="text-gray-500"
            size={16}
          />
```

src/components/Files/FileItemContextMenu.tsx
```diff
@@ -0,0 +1,73 @@
...
+}
+
+export const FileItemContextMenu = ({
+  file,
+  className,
+  onDelete,
+}: ContextMenuProps) => {
+  const { t } = useTranslation('sidebar');
+  return (
+    <>
```
IlyaBondar: You don't need to use a fragment here, because only one tag/component `Menu` inside

src/components/Chat/ChatInput/AttachButton.tsx
```diff
@@ -0,0 +1,126 @@
...
+            }
+            onClick={handleAttachFromComputer}
+          />
+        </Menu>
+      </div>
+      {isSelectFilesDialogOpened && (
+        <FileSelect
+          isOpen
+          // TODO: take this from model settings
+          allowedTypes={['.dmg', '.png', '.pdf']}
```
IlyaBondar: `['.dmg', '.png', '.pdf']` is used twice in this file and will be new array each time (rerender). Use constant outside component or `createSelector` to memoize it

src/components/Chat/ChatInput/AttachButton.tsx
```diff
@@ -0,0 +1,126 @@
...
+          onClose={(result) => {
```
IlyaBondar: should be wrapped in `useCallback` and moved out of markup

src/components/Chat/ChatInput/AttachButton.tsx
```diff
@@ -0,0 +1,126 @@
...
+            setIsSelectFilesDialogOpened(false);
+          }}
+        />
+      )}
+      {isPreUploadDialogOpened && (
+        <PreUploadDialog
+          isOpen
+          // TODO: take this from model settings
+          allowedTypes={['.dmg', '.png', '.pdf']}
+          onUploadFiles={(
```
IlyaBondar: should be wrapped in `useCallback` and moved out of markup

src/components/Chat/ChatInput/ChatInput.tsx
```diff
@@ -160,13 +167,30 @@ export const ChatInput = ({
...
+            .filter(
+              (file) => file.status !== 'FAILED' && file.status !== 'UPLOADING',
+            )
+            .map((file) => ({
+              type: file.contentType,
+              title: file.name,
+              url: encodeURI(file.absolutePath + '/' + file.name),
+            })),
+        },
+      }),
```
IlyaBondar: this block isn't readable. As I understood correctly it should be like this:
```suggestion
      custom_content: files.length > 0 ? {
          attachments: files
            .filter(
              (file) => file.status !== 'FAILED' && file.status !== 'UPLOADING',
            )
            .map((file) => ({
              type: file.contentType,
              title: file.name,
              url: encodeURI(`${file.absolutePath}/${file.name}`),
            })),
        },
      } : undefined,
```
but even better to move this code into util-method `getCustomContent(files)` outside this component/file and use:
``` javascript
custom_content: getCustomContent(files)
```

src/components/Chat/ChatInput/ChatInputAttachments.tsx
```diff
@@ -0,0 +1,86 @@
...
+                </div>
+              )}
+            </div>
+            <div className="flex gap-3">
+              {file.status === 'FAILED' && (
+                <button onClick={handleRetry(file.id)}>
+                  <IconReload className="shrink-0 text-gray-500" size={18} />
+                </button>
+              )}
+              <button onClick={handleUnselectFile(file.id)}>
```
IlyaBondar: `handleRetry` and `handleUnselectFile` are memoized by `useCallback`, **BUT** `handleRetry(file.id)` and `handleUnselectFile(file.id)` return each time new functions!

it's better to move these handlers and markup in `ChatInputAttachment` and use here
```
{selectedFiles.map(file => (
        <ChatInputAttachment
          key={file.id} file={file} />
 )}
 ```
or even better if `ChatInputAttachment` will read file by id from state
```
{selectedFilesIds.map(fileId => (
        <ChatInputAttachment
          key={fileId} fileId={fileId} />
 )}
 ```

src/components/Chat/ChatInput/PromptDialog.tsx
```diff
@@ -17,8 +17,8 @@ import { onBlur } from '@/src/utils/app/style-helpers';
 
 import { Prompt } from '@/src/types/prompt';
 
-import XMark from '../../../public/images/icons/xmark.svg';
-import EmptyRequiredInputMessage from '../Common/EmptyRequiredInputMessage';
+import XMark from '../../../../public/images/icons/xmark.svg';
```
IlyaBondar: It would be nice to use `IconX` from `@tabler/icons-react` instead

src/components/Spinner/Spinner.tsx
```diff
@@ -1,32 +1,20 @@
+import classNames from 'classnames';
+
 interface Props {
-  size?: string;
+  size?: number;
   className?: string;
 }
 
-const Spinner = ({ size = '1em', className = '' }: Props) => {
+const Spinner = ({ size = 16, className = '' }: Props) => {
```
IlyaBondar: Can we move this component into `Common` folder?

src/components/Chat/ChatMessage.tsx
```diff
@@ -255,8 +255,15 @@ export const ChatMessage: FC<Props> = memo(
...
-                      {message.content}
+                    <div className="mr-2 flex w-full flex-col gap-5">
+                      <div className="prose flex-1 whitespace-pre-wrap dark:prose-invert">
+                        {message.content}
+                      </div>
+                      {!!message.custom_content?.attachments?.length && (
+                        <MessageAttachments
+                          attachments={message.custom_content.attachments}
+                        />
+                      )}
```
IlyaBondar: this condition `!!message.custom_content?.attachments?.length` should be moved inside `MessageAttachments` (return null if `attachments` are empty), because this component is responsible for checking and displaying `attachments`
```suggestion
                      <MessageAttachments
                          attachments={message.custom_content?.attachments}
                        />
```

src/components/Chat/MessageAttachment.tsx
```diff
@@ -48,45 +53,53 @@ export const MessageAttachment = ({ attachment, isInner }: Props) => {
...
+            </a>
+          )}
         </button>
       </div>
-      {(attachment.data || attachment.url) && isOpened && (
+      {isOpenable && attachment.data && isOpened && (
         <div
           className={`relative mt-2 h-auto w-full overflow-hidden p-3 pt-4 text-sm duration-200`}
         >
           {imageTypes.includes(attachment.type) ? (
```
IlyaBondar: it would be nice to save `imageTypes` in constant outside the component(new array each render) and convert it to `Set` (O(n)->O(1))
```
const imageTypes: ImageMIMEType[] = new Set(['image/jpeg', 'image/png']);
```
and use
```suggestion
          {imageTypes.has(attachment.type) ? (
```

src/components/Common/FolderContextMenu.tsx
```diff
@@ -1,25 +1,41 @@
...
 import { HighlightColor } from '@/src/types/components';
 
 import DotsIcon from '../../../public/images/icons/dots-vertical.svg';
 import { Menu, MenuItem } from './DropdownMenu';
 
 interface FolderContextMenuProps {
-  onDelete: MouseEventHandler<unknown>;
-  onRename: MouseEventHandler<unknown>;
+  onDelete?: MouseEventHandler<unknown> | undefined;
+  onRename?: MouseEventHandler<unknown> | undefined;
```
IlyaBondar: `| undefined` is extra here

src/components/Files/FileFolderSelect.tsx
```diff
@@ -0,0 +1,278 @@
...
+  const [selectedFolderId, setSelectedFolderId] = useState<string | undefined>(
+    selectedFolderName,
+  );
+
+  useEffect(() => {
+    if (isOpen) {
+      dispatch(FilesActions.getFolders({}));
+      openedFoldersIds.forEach((folderId) => {
+        dispatch(FilesActions.getFolders({ path: folderId }));
+      });
```
IlyaBondar: executing `dispatch` in cycle isn't good practice.
Add some action which can work with array of ids:
```suggestion
      dispatch(FilesActions.getFolderList(openedFoldersIds));
```
and what does `dispatch(FilesActions.getFolders({}));` do? Can it be included into `getFolderList`? e.g. with flag `uploadRoot`

src/components/Files/FileFolderSelect.tsx
```diff
@@ -0,0 +1,278 @@
...
+                <IconX className="text-gray-500" />
+              </button>
+              <div className="flex flex-col gap-2 overflow-auto">
+                <div className="flex justify-between">
+                  <h2 id={headingId} className="text-base font-semibold">
+                    {t('Select folder')}
+                  </h2>
+                </div>
+                {folders.length === 0 &&
+                ['LOADING', undefined].includes(foldersStatus) ? (
```
IlyaBondar: Move `['LOADING', undefined]` in constant outside the component to not create array each time and wrap into `Set`
```
const loadingStatuses = new Set(['LOADING', undefined]);
```
and use
```suggestion
                loadingStatuses.has(foldersStatus) ? (
```

src/components/Files/FileFolderSelect.tsx
```diff
@@ -0,0 +1,278 @@
...
+                                  <div key={folder.id}>
+                                    <Folder
+                                      searchTerm={searchQuery}
+                                      currentFolder={folder}
+                                      allFolders={folders}
+                                      highlightColor="blue"
+                                      highlightedFolders={
+                                        selectedFolderId
+                                          ? [selectedFolderId]
+                                          : []
```
IlyaBondar: new array each time - rerender

src/components/Files/FileItem.tsx
```diff
@@ -0,0 +1,129 @@
...
+import classNames from 'classnames';
+
+import { DialFile } from '@/src/types/files';
+
+import { FilesSelectors } from '@/src/store/files/files.reducers';
+import { useAppSelector } from '@/src/store/hooks';
+
+import { FileItemContextMenu } from './FileItemContextMenu';
+
+export type FileItemEventIds = 'cancel' | 'retry' | 'toggle' | 'remove';
```
IlyaBondar: it would be nice to use `enum` instead and in all places for these values

src/components/Files/FileItem.tsx
```diff
@@ -0,0 +1,129 @@
...
+              style={{ width: `${item.percent}%` }}
+            ></div>
+          </div>
+        )}
+        {item.status === 'FAILED' && (
+          <button onClick={handleRetry}>
+            <IconReload className="shrink-0 text-gray-500" size={18} />
+          </button>
+        )}
+        {item.status && ['UPLOADING', 'FAILED'].includes(item.status) ? (
```
IlyaBondar: `['UPLOADING', 'FAILED']` should be moved in contants outside component(new array each time) and wrapped in `new Set` 
```suggestion
        {item.status && statusesAllowCancel.has(item.status) ? (
```

IlyaBondar (changes requested):
reviewed 18 out of 47 files

src/components/Chat/ChatInput/ChatInput.tsx
mikitabut: Now it's needed :)

src/components/Chat/ChatInput/ChatInputAttachments.tsx
mikitabut: @IlyaBondar I don't think this is recommended - just possibilities how to use. It's anyway resolves to false and being filtered from resulting list of values

src/components/Chat/ChatInput/ChatInputAttachments.tsx
mikitabut: It's OK but according to this - "even better if ChatInputAttachment will read file by id from state'. Why it's better? I can just pass the object and will get no any problems, but otherwise this new component will be hardly connected with what we have in store. As from my point of view it's not need, because if we will have same file component in other places but with other data it will require return it back

So I will use your first suggested change

mikitabut:
/deploy-review
>Deployment status: success

src/components/Chat/ChatInput/AttachButton.tsx
mikitabut: It was redone with listing param support

src/components/Folder/Folder.tsx
mikitabut: Actually I'm not sure that this is working, because tailwind is checking for colors by some pattern and if we will not use them directly I'm not sure that this will work
Moreover we have some linters and classnames needed for them correctly working - they show is class exists or not

IlyaBondar: @mikitabut, this method will work correctly as `classNames` - no difference.
Intellisence also will work:
image
image

mikitabut:
/deploy-review
>Deployment status: success

src/components/Sidebar/BetweenFoldersLine.tsx
```diff
@@ -62,8 +63,13 @@ export const BetweenFoldersLine = ({
...
 
-  const highlightColorBg =
-    highlightColor === 'green' ? 'bg-green/60' : 'bg-violet/60';
+  const highlightColorBg = classNames(
+    highlightColor === 'green'
+      ? 'bg-green/60'
+      : highlightColor === 'violet'
+      ? 'bg-violet/60'
+      : 'bg-blue-500/60',
+  );
```
IlyaBondar: Can wrap into `getHighlightColor`

src/components/Files/PreUploadModal.tsx
```diff
@@ -0,0 +1,413 @@
...
+      }, []);
+      if (incorrectSizeFiles.length > 0) {
+        setErrorMessage(
+          (oldMessage) =>
+            oldMessage +
+            '\n' +
+            t(
+              `Max file size up to 512 Mb. Next files haven't been uploaded: `,
+            ) +
+            incorrectSizeFiles.join(', '),
```
IlyaBondar: Can we use string interpolation here?

src/components/Files/PreUploadModal.tsx
```diff
@@ -0,0 +1,413 @@
...
+      if (incorrectTypeFiles.length > 0) {
+        setErrorMessage(
+          (oldMessage) =>
+            oldMessage +
+            '\n' +
+            t(
+              `Supported types: {{allowedTypes}}. Next files haven't been uploaded: `,
+              { allowedTypes: allowedTypes.join(', ') },
+            ) +
+            incorrectTypeFiles.join(', '),
```
IlyaBondar: Can we use string interpolation here?

src/components/Files/PreUploadModal.tsx
```diff
@@ -0,0 +1,413 @@
...
+      }
+    },
+    [allowedTypes, t],
+  );
+
+  const handleUpload = useCallback(() => {
+    if (attachments.length + selectedFiles.length > 10) {
+      setErrorMessage(
+        t(
+          `Maximum allowed attachments number is {{maxAttachmentsAmount}}. With your uploadings amount will be {{selectedAttachmentsAmount}}`,
```
IlyaBondar, line 165: How are we going to use such long text as key for others locales (i18n)?

src/components/Common/FolderContextMenu.tsx
```diff
@@ -1,69 +1,91 @@
...
       type="contextMenu"
       className="justify-self-end"
       trigger={
-        <DotsIcon
-          className="rotate-90 text-gray-500"
-          width={18}
-          height={18}
+        <IconDots
+          className="text-gray-500"
           size={18}
```
IlyaBondar: if we don't want to see difference with previous icon you need to decrease size
```suggestion
          size={16}
```

IlyaBondar (dismissed):

src/components/Files/PreUploadModal.tsx
mikitabut: We don't have any other locales for now, so when we will work on that we will need to refactor all such texts and replace them with keys, but for now we are not doing that for simplicity

IlyaBondar (approved):

Issue #119, opened 2023-11-14
Unnecessary warning about code correctness, security and licensing appears in some specific cases


VolhaBazhkova:
Case1: Please see the screenshot -> there is no code block => no warning should appear. Use the file to reproduce the problem.
[chatbot_ui_conversation_case1.json](https://github.com/epam/ai-dial-chat/files/13349059/chatbot_ui_conversation_case1.json)

![image](https://github.com/epam/ai-dial-chat/assets/111180458/9f56c427-ff93-4d59-af73-3d2b2023586d)

Case2: 
[chatbot_ui_conversation_case2.json](https://github.com/epam/ai-dial-chat/files/13349069/chatbot_ui_conversation_case2.json)

![image](https://github.com/epam/ai-dial-chat/assets/111180458/9e78c5ac-25c4-4d59-8ba6-3146ff71c4d9)

Case3: 
[chatbot_ui_conversation_case3.json](https://github.com/epam/ai-dial-chat/files/13349072/chatbot_ui_conversation_case3.json)
![image](https://github.com/epam/ai-dial-chat/assets/111180458/b83c17ae-9a55-47b8-9edc-8250c04b5f3b)

VolhaBazhkova:
Now the warning message doesn't appear in the cases described in the ticket. Though there still could be an unnecessary warning if 4 spaces (or tab) are received in response. Example, 
[chatbot_ui_conversation_10-3__4_ (1).json](https://github.com/epam/ai-dial-chat/files/13449863/chatbot_ui_conversation_10-3__4_.1.json)
![image](https://github.com/epam/ai-dial-chat/assets/111180458/8e200542-5cc0-4e70-a79a-b993da5cee48)

VolhaBazhkova:
One more example 

[epam_ai_dial_-_development_chat_conversation_5-6.json](https://github.com/user-attachments/files/20059544/epam_ai_dial_-_development_chat_conversation_5-6.json)

PR #200, opened 2023-11-22
fix: correct RegEx for the code detection

denys-kolomiitsev:
Changed RegEx for the code detection for the worrning in the message.

Issue #119


denys-kolomiitsev:
/deploy-review
>Deployment status: success

mikitabut (approved):

PR #204, opened 2023-11-23
chore: use enums and css helper 

IlyaBondar:
### Description
- use enums/constants
- use CSS-helper `getByHighlightColor`
- fix dependencies


IlyaBondar:
/deploy-review
>Deployment status: success

.eslintrc.json
```diff
@@ -26,7 +26,10 @@
       { "argsIgnorePattern": "^_" }
     ],
     "no-restricted-imports": ["error", { "paths": ["react-i18next"] }],
-    "tailwindcss/no-custom-classname": "error",
+    "tailwindcss/no-custom-classname": ["error", { "callees": [
```
mikitabut: This is duplication of line 16 - use it there because it will affect not only 'no-custom-classname' but other rules also

IlyaBondar:
/deploy-review
>Deployment status: success

src/types/openai.ts
```diff
@@ -1,7 +1,9 @@
-export type OpenAIEntityModelType = 'model';
-export type OpenAIEntityApplicationType = 'application';
-export type OpenAIEntityAssistantType = 'assistant';
-export type OpenAIEntityAddonType = 'addon';
+import { EntityType } from './common';
+
+export type OpenAIEntityModelType = EntityType.Model;
```
mikitabut: Do we need this if you already have enum?

.eslintrc.json
IlyaBondar: Hm, you are right we need to apply all rules(order, no-custom and etc.) except `tailwindcss/no-contradicting-classname`(not so smart that we are choosing between) for `getByHighlightColor`

e2e/src/testData/conversationHistory/conversationData.ts
```diff
@@ -1,7 +1,9 @@
 import { DateUtil } from '@/e2e/src/utils/dateUtil';
 import { GeneratorUtil } from '@/e2e/src/utils/generatorUtil';
 
-import { Conversation, Message, Stage } from '@/src/types/chat';
+
```
mikitabut: as for me too much empty rows :)

mikitabut (dismissed):

mikitabut (approved):

Issue #207, opened 2023-11-23
Files attachment corrections


DaryaKarniyenka:
Parent ticket - https://github.com/epam/ai-dial-chat/issues/80

- We should add blue background for files and folders on hover. 
Also we should show checkbox while hovering on the file row, now hover works only directly on icon. 
Also add blue line from the left side for the select state.

- It's impossible to upload file via folder dots. We should add this function 

- Incorrect text color for button in light mode 

![image](https://github.com/epam/ai-dial-chat/assets/151838896/9f3acac9-d7ec-4c51-9232-7a853aa1a816)

- Incorrect hover color in dropdown. It should be blue. Also row height and icon size should be smaller.

![image](https://github.com/epam/ai-dial-chat/assets/151838896/0f272821-5e75-4a86-995c-a512c1d7a53b)

- Error style should be different (https://www.figma.com/file/lwL5jneveZczAj5RLcLFH2/DIAL-Chat?type=design&node-id=96%3A3777&mode=design&t=5jqonWtraLYnRNBL-1)

![image](https://github.com/epam/ai-dial-chat/assets/151838896/d49b315f-8ada-4e2a-aae5-6316e4ea2811)

- Add an icon for this error

![image](https://github.com/epam/ai-dial-chat/assets/151838896/60847cb7-e969-49f0-9154-66c7e578d0db)

- "Add Folder" button should have the same hover effect as side menu icons in the bottom. Also this icon is not centred 

![image](https://github.com/epam/ai-dial-chat/assets/151838896/6dc9771e-0218-4303-8bcf-8eca383f9b7a)

Also it's very inconvenient that it's impossible to rename, move and delete folders/files right now. Hope we will add these functions soon.

mikitabut:
Point 2 - will be fixed in https://github.com/epam/ai-dial-chat/issues/211

VolhaBazhkova:
Verified on staging successfully. There are some comments:
- It's impossible to upload file via folder dots. We should add this function. Fill be fixed in #230 
- folder background highlight is not done. It seems that it should be implemented at the same time with #91

PR #214, opened 2023-11-23
fix: fix design issues

mikitabut:
Refs: #207

### DIFF
--- a/src/components/Files/FileFolderSelect.tsx
+++ b/src/components/Files/FileFolderSelect.tsx
@@ -141,8 +141,12 @@ export const FileFolderSelect = ({
   );
   const handleRenameFolder = useCallback(
     (newName: string, folderId: string) => {
+      const renamingFolder = folders.find((folder) => folder.id === folderId);
       const folderWithSameName = folders.find(
-        (folder) => folder.name === newName && folderId !== folder.id,
+        (folder) =>
+          folder.name === newName &&
+          folderId !== folder.id &&
+          folder.folderId === renamingFolder?.folderId,
       );
 
       if (folderWithSameName) {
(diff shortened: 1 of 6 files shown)


src/components/Files/FileFolderSelect.tsx
```diff
@@ -141,8 +141,12 @@ export const FileFolderSelect = ({
   );
   const handleRenameFolder = useCallback(
     (newName: string, folderId: string) => {
+      const renamingFolder = folders.find((folder) => folder.id === folderId);
       const folderWithSameName = folders.find(
-        (folder) => folder.name === newName && folderId !== folder.id,
+        (folder) =>
+          folder.name === newName &&
```
IlyaBondar: SHould we use `trim` here?

IlyaBondar (dismissed):

denys-kolomiitsev (approved):

src/components/Files/FileFolderSelect.tsx
```diff
@@ -141,8 +141,12 @@ export const FileFolderSelect = ({
   );
   const handleRenameFolder = useCallback(
     (newName: string, folderId: string) => {
+      const renamingFolder = folders.find((folder) => folder.id === folderId);
       const folderWithSameName = folders.find(
-        (folder) => folder.name === newName && folderId !== folder.id,
+        (folder) =>
+          folder.name === newName.trim() &&
```
IlyaBondar, line 147: Do we want to case-insensitive comparison?

mikitabut: Nope, I don't think that we need case-insensitive comparison
Sometimes I have such folders like `super folder !` and `SUPER folder!`, so for me it will be not so good UX

IlyaBondar (approved):

Issue #86, opened 2023-11-13
Chat should be automatically renamed if the conversation is started from the beginning after clearing all messages


VolhaBazhkova:
Chat should be automatically renamed if the conversation is started from the beginning after clearing all messages. Mainly this issue is inconvenient in **overlay** functionality, as there is no way to rename the chat.

1. Create a chat and send the first message -\> the chat is named automatically, where the 1st message = the name
2. In chat header click on "Clear all messages" button; confirm -\> messages from the chat are removed (ok); chat settings appears (ok); chat name stays the same -\> **issue1?**
3. Send new first message -\> chat name is not changed -\> **issue2**

**Expected result:** After "Clear all messages" action, the chat should be renamed automatically to the new first message if previous name was also named automatically. Please do nothing if the chat was renamed manually.

**To be discussed** what is the best approach: should we rename the chat to New conversation on the 2nd step or only on the 3rd step

Note: If all the messages were deleted one by one, expected result is the same

VolhaBazhkova:
The fix was verified on staging environment successfully. 
Now the behaviour is the following:
1. Create a chat and send a message -> chat is named to the 1st message
2. Clear all the messages on the button in chat header or deleting user-messages one by one -> all the messages are deleted, chat is named as it was in the 1st step
3. Send new message -> chat is renamed to this new message automatically -> ok! fixed!

Note: if the chat was renamed manually, it is not renamed automatically on the 3rd step

Reported #225

PR #216, opened 2023-11-23
fix: automatically rename chat after clearing all messages

denys-kolomiitsev:
Implemented renaming conversation to the default name in case of clearing all messages in it.
Moreover, added flag to not rename if it was renamed by user.

Issue refs: #86

### DIFF
--- a/src/components/Chat/Chat.tsx
+++ b/src/components/Chat/Chat.tsx
@@ -271,13 +271,16 @@ export const Chat = memo(() => {
         confirm(t<string>('Are you sure you want to clear all messages?')) &&
         conversation
       ) {
-        const { messages } = conversation;
+        const { messages, isNameChanged, name } = conversation;
+        const newConversationName = isNameChanged
+          ? name
+          : DEFAULT_CONVERSATION_NAME;
 
         dispatch(
           ConversationsActions.updateConversation({
             id: conversation.id,
             values: {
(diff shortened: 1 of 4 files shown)


denys-kolomiitsev:
/deploy-review
>Deployment status: success

IlyaBondar (dismissed):

src/components/Chat/Chat.tsx
```diff
@@ -269,12 +272,16 @@ export const Chat = memo(() => {
...
+        const { messages, isNameChanged, name } = conversation;
+        const newConversationName = isNameChanged
+          ? name
+          : DEFAULT_CONVERSATION_NAME;
 
         dispatch(
           ConversationsActions.updateConversation({
             id: conversation.id,
             values: {
+              name: newConversationName,
```
mikitabut: You should do this not here but where you have changing name for empty chats, because you are not covered case when user will delete messages one by one

mikitabut (changes requested):

src/components/Chat/Chat.tsx
denys-kolomiitsev: Thank you for pointing out this case.
Moved condition to the sendMessageEpic

denys-kolomiitsev:
/deploy-review
>Deployment status: success

mikitabut (approved):

PR #217, opened 2023-11-24
fix: fix attachments design issues

mikitabut:
Refs: #207

### DIFF
--- a/src/components/Files/FileFolderSelect.tsx
+++ b/src/components/Files/FileFolderSelect.tsx
@@ -141,8 +141,12 @@ export const FileFolderSelect = ({
   );
   const handleRenameFolder = useCallback(
     (newName: string, folderId: string) => {
+      const renamingFolder = folders.find((folder) => folder.id === folderId);
       const folderWithSameName = folders.find(
-        (folder) => folder.name === newName && folderId !== folder.id,
+        (folder) =>
+          folder.name === newName &&
+          folderId !== folder.id &&
+          folder.folderId === renamingFolder?.folderId,
       );
 
       if (folderWithSameName) {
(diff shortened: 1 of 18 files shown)


src/components/Files/FileFolderSelect.tsx
```diff
@@ -141,16 +141,17 @@ export const FileFolderSelect = ({
   );
   const handleRenameFolder = useCallback(
     (newName: string, folderId: string) => {
+      const renamingFolder = folders.find((folder) => folder.id === folderId);
       const folderWithSameName = folders.find(
-        (folder) => folder.name === newName && folderId !== folder.id,
+        (folder) =>
+          folder.name === newName.trim() &&
```
IlyaBondar: Don't we still want to use case-insensitive comparison?

src/components/Files/PreUploadModal.tsx
```diff
@@ -317,17 +318,14 @@ export const PreUploadDialog = ({
                   </div>
                   <p id={descriptionId}>
                     {t(
-                      'Max file size up to 1 GB. Supported types: {{allowedTypes}}.',
+                      'Max file size up to 512 Mb. Supported types: {{allowedTypes}}.',
```
IlyaBondar, line 321: Do we really want to support such big files (512 Mb)?

IlyaBondar (dismissed):

src/components/Files/FileFolderSelect.tsx
mikitabut: Yes

src/components/Files/PreUploadModal.tsx
mikitabut: Yes)

IlyaBondar (approved):

Issue #211, opened 2023-11-23
[Part 2] Implement attachments


mikitabut:
According to https://github.com/epam/ai-dial-chat/issues/80 we have some leftovers which will be fixed in scope of current ticket

1. Separate Modal to manage files which opened from left side bar
2. Map mime types to extensions on ui
3. Edit message

VolhaBazhkova:
Verified on staging environment. Reported #278

PR #222, opened 2023-11-24
feat: additional features for attachments

mikitabut:
https://github.com/epam/ai-dial-chat/issues/211

Next cases covered:
Separate Modal to manage files which opened from left side bar
Map mime types to extensions on UI
Bulk remove
Bulk download


mikitabut:
/deploy-review
>Deployment status: success

src/components/Files/PreUploadModal.tsx
```diff
@@ -98,25 +110,20 @@ export const PreUploadDialog = ({
...
+        512 * bytesInMb,
+      ).map((file) => file.name);
+      const incorrectTypeFiles: string[] = getFilesWithInvalidFileType(
+        files,
+        allowedTypes,
+      ).map((file) => file.name);
+      const filteredFiles = files.filter(
+        (file) =>
+          !incorrectSizeFiles.includes(file.name) &&
+          !incorrectTypeFiles.includes(file.name),
```
IlyaBondar, line 124: wrap `incorrectSizeFiles` and `incorrectTypeFiles` into `new Set` to search with O(1)
```suggestion
          !incorrectSizeFiles.has(file.name) &&
          !incorrectTypeFiles.has(file.name),
```

src/pages/api/files/index.ts
```diff
@@ -85,8 +85,8 @@ async function handleGetRequest(
 ) {
   const { path } = req.query as { path: string };
   const url = `${process.env.OPENAI_API_HOST}/v1/files${
-    path && `/${encodeURI(path)}`
-  }?path=absolute&purpose=content`;
+    path && encodeURI(path)
```
IlyaBondar, line 88: can path be empty? And if yes we will inject "false" into url in this case

IlyaBondar (approved):

src/pages/api/files/index.ts
mikitabut: By design it can't be empty, because every file has path and it's passing to URL
But I will add some checks

Issue #82, opened 2023-11-13
Disable Send icon if there is empty response for Models/Assistants/Applications


VolhaBazhkova:
1. Create any chat based on any Application/Model/Assistant
2. Type and send any message
3. Click on Stop generating immediately -> empty row is in response
4. Hover over Send icon in message box -> send action is available -> **bug**

   **Expected result**: Send icon should be disabled, tooltip-explanation should be shown "Please regenerate response to continue working with chat" and mouse arrow should be changed to "stop/forbidden" icon ![image](https://github.com/epam/ai-dial-chat/assets/111180458/a8262b5c-3c6c-4c38-baff-003de1a011b5)

![image](https://github.com/epam/ai-dial-chat/assets/111180458/cceebf69-80fa-4500-bd4f-bb006ec92645)

VolhaBazhkova:
Verified on staging env. successfully.
![image](https://github.com/epam/ai-dial-chat/assets/111180458/38e91635-8b06-41c9-8ced-fca6e49b0dba)

PR #246, opened 2023-11-28
fix: fix send message by enter when message is empty

mikitabut:
#80
#82

### DIFF
--- a/src/components/Chat/ChatInput/ChatInputMessage.tsx
+++ b/src/components/Chat/ChatInput/ChatInputMessage.tsx
@@ -89,7 +89,11 @@ export const ChatInputMessage = ({
       prompt.name.toLowerCase().includes(promptInputValue.toLowerCase()),
     ),
   );
-
+  const isInputEmpty = useMemo(() => {
+    return (
+      (!content || content.trim().length === 0) && selectedFiles.length === 0
+    );
+  }, [content, selectedFiles.length]);
   const maxLength = useMemo(() => {
     const maxLengthArray = selectedConversations.map(
       ({ model }) =>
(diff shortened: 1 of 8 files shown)


src/components/Chat/ChatInput/ChatInputMessage.tsx
```diff
@@ -138,14 +142,14 @@ export const ChatInputMessage = ({
       return;
     }
 
-    if (!content) {
+    if (isInputEmpty) {
       alert(t('Please enter a message'));
       return;
     }
```
IlyaBondar: I think we don't want to show this alert and should do just like this:
```javascript
   if (messageIsStreaming || isInputEmpty) {
        return;
   }
```
plus in `SendMessageButton` we need to add check for error cases to avoid sending on Enter click:
```javascript
onClick={!disabled ? handleSend : undefined}
```
where
```javascript
const disabled = messageIsStreaming || isModelsLoading || isShowDisabled;
```

IlyaBondar (changes requested):

IlyaBondar:
\deploy-review

src/components/Chat/ChatInput/ChatInputMessage.tsx
mikitabut: Done slightly in different way

mikitabut:
/deploy-review

>Deployment status: success

src/components/Chat/ChatInput/SendMessageButton.tsx
```diff
@@ -45,56 +38,39 @@ const SendIconTooltip = ({
...
+interface Props {
+  handleSend: () => void;
+  isDisabled: boolean;
+  disabledTooltip?: string;
+}
+
+export const SendMessageButton = ({
+  handleSend,
+  isDisabled,
+  disabledTooltip,
```
IlyaBondar: we display tooltip even when not disabled. I would rename to `tooltip`

IlyaBondar (dismissed):

src/components/Chat/ChatInput/SendMessageButton.tsx
mikitabut: fixed

IlyaBondar (approved):

PR #285, opened 2023-12-04
fix: change header from event stream to just octet stream

mikitabut:


mikitabut:
/deploy-review
>Deployment status: success

mikitabut:
/deploy-review
>Deployment status: success

IlyaBondar (approved):

Issue #300, opened 2023-12-07
Need to reorder coversations and prompts settings


IlyaBondar:
#### Changes:
- change "create folder" and "delete all" positions
- hide "export" if no prompts/folders
- show "delete all" if have only folders

conversations:
![image](https://github.com/epam/ai-dial-chat/assets/20163971/eb349091-c5f8-44ef-a6d7-f89e8dcf39f4)

prompts(empty):
![image](https://github.com/epam/ai-dial-chat/assets/20163971/2f7091b5-ec82-4674-9ef1-e290477bba68)
prompts(not empty):
![image](https://github.com/epam/ai-dial-chat/assets/20163971/dc8fd5ef-ef70-4a07-b868-c4641df3b1bf)

VolhaBazhkova:
Verified on staging successfully.

PR #301, opened 2023-12-07
chore: reorder coversations and prompts settings (#300)

IlyaBondar:
### Description
- Issue #300 

#### Changes:
- change "create folder" and "delete all" positions
- hide "export" if no prompts/folders
- show "delete all" if have only folders

conversations:
image

prompts(empty):
image
prompts(not empty):
image


IlyaBondar:
/deploy-review
>Deployment status: success

denys-kolomiitsev (approved):

Issue #143, opened 2023-11-14
Selected but not saved theme is set by default when user opens Settings


VolhaBazhkova:
1. Clear site data
2. Click on Settings -\> Dark mode is set by default -\> ok
3. Select Light mode (don't click on Save)
4. Click on any area outside the pop -\> Settings is closed, the theme is still dark mode -\> ok
5. Click on Setting again -\> Light mode is set by default -\> bug

**Expected result:** Currently used mode should be shown in theme list when the pop-up is opened (as it's selected now).

![image](https://github.com/epam/ai-dial-chat/assets/111180458/9d6dfd26-6995-40d2-a415-d278a2913734)

VolhaBazhkova:
Verified on staging successfully

Issue #83, opened 2023-11-13
Need to add an option in Settings to allow to use entire screen width for chat messages and message box


VolhaBazhkova:
We need to add an option in Settings to allow to use entire screen width for chat messages and message box. By default the option should be off. Please find the design here (the link is available for DIAL dev team only): https://www.figma.com/file/lwL5jneveZczAj5RLcLFH2/DIAL-Chat?type=design&node-id=1479-6708&mode=design&t=bvYoJf5FCesvhuAZ-4 

![image](https://github.com/epam/ai-dial-chat/assets/111180458/aea13edc-bd0d-4860-b938-f7de9c3b52b8)

Comment from user (246 internal ticket): `Please consider increase the conversation width, it currently only occupy less than 50% approx. of the visible area width. This drawback becomes quite outstanding when the conversation include table format content, when each column becomes very narrow, it's difficult to read.`

VolhaBazhkova:
Please take care of #143 #144 
Can we fix them at the same time?

VolhaBazhkova:
@denys-kolomiitsev return the feature:


![Image](https://github.com/epam/ai-dial-chat/assets/111180458/d653e681-edf7-48db-ae56-eb17ce5d927b)

denys-kolomiitsev:
The width of the conversation name container depends on the conversation name length. Now default renaming after the first message cuts name at 160th character.
![image](https://github.com/epam/ai-dial-chat/assets/143205282/905c9c3b-a6c1-4576-8887-a4a517e28454)

If manualy rename conversation it will widen to full width.
![image](https://github.com/epam/ai-dial-chat/assets/143205282/8a311d97-2717-4cc2-9136-8f6c8cb11d6f)

TBD: should we increase characters number of the default conversation name.

VolhaBazhkova:
Verified on staging successfully

PR #346, opened 2023-12-15
feat: added  an full screen chat width option(#83, #143)

denys-kolomiitsev:
-  #83 
- #143 

image

Refs: #83, #143


denys-kolomiitsev:
/deploy-review
>Deployment status: success

src/components/Settings/ToggleFullWidth.tsx
mikitabut, line 1: Please create generic component to being able to use not only for settings but for other stuff such switcher too

src/components/Settings/SettingDialog.tsx
```diff
@@ -123,3 +125,5 @@ export const SettingDialog: FC<Props> = ({ open, onClose }) => {
     </FloatingPortal>
   );
 };
+
+export default SettingDialog;
```
mikitabut: Is it really needed change? Now it's harder to import settings dialog if I will need it somewhere else

denys-kolomiitsev:
/deploy-review
>Deployment status: success

src/components/Settings/SettingDialog.tsx
```diff
@@ -17,27 +10,42 @@ import { useAppDispatch, useAppSelector } from '@/src/store/hooks';
...
+
   useEffect(() => {
     setLocalTheme(theme);
   }, [theme]);
 
+  useEffect(() => {
+    setIsChatFullWidthLocal(isChatFullWidth);
+  }, [isChatFullWidth]);
+
   useEffect(() => {
```
mikitabut, line 50: Is it possible to use useDismiss and other stuff from floating UI?

mikitabut (changes requested):

src/components/Settings/ToggleFullWidth.tsx
```diff
@@ -0,0 +1,53 @@
...
+        type="checkbox"
+        onChange={handleSwitch}
+        id="toggle"
+        className="sr-only"
+        checked={isOn}
+      />
+
+      <div
+        className={classNames(
+          'flex h-[22px] w-[50px] shrink-0 cursor-pointer items-center  gap-1 rounded-2xl p-1.5 transition-all duration-200',
```
mikitabut: maybe rounded-full better than rounded-2xl?

mikitabut: Having fixed height limited us from using it with other labels - suggest to get rid of this fixed width

src/components/Settings/ToggleFullWidth.tsx
```diff
@@ -0,0 +1,53 @@
...
+}: ToggleFullWidthProps) => {
+  const { t } = useTranslation(Translation.Settings);
+
+  const switchBackground = isOn
+    ? 'bg-blue-500 dark:bg-blue-500'
+    : 'bg-gray-400 dark:bg-gray-600';
+
+  return (
+    <label htmlFor="toggle" className="flex w-full items-center gap-5">
+      <span className="w-[120px]">{t('Full width chat')}</span>
```
mikitabut: Why do you need to have fixed width here?

src/components/Settings/SettingDialog.tsx
```diff
@@ -17,27 +10,42 @@ import { useAppDispatch, useAppSelector } from '@/src/store/hooks';
...
+import { ThemeSelect } from './ThemeSelect';
+import { ToggleFullWidth } from './ToggleFullWidth';
 
 interface Props {
   open: boolean;
   onClose: () => void;
 }
 
-export const SettingDialog: FC<Props> = ({ open, onClose }) => {
+const SettingDialog: FC<Props> = ({ open, onClose }) => {
```
IlyaBondar: it would be better to use default import right here
```suggestion
export default function SettingDialog({ open, onClose }: Props) {
```
https://react.dev/learn/importing-and-exporting-components#default-vs-named-exports

src/components/Settings/SettingDialog.tsx
denys-kolomiitsev: It is possible, but I think this is from scope of other issue.

src/components/Settings/ToggleFullWidth.tsx
denys-kolomiitsev: Made it more generic for reusability.

src/components/Settings/SettingDialog.tsx
denys-kolomiitsev: It is  not really needed. However, it is common practice:
 https://react.dev/learn/importing-and-exporting-components#default-vs-named-exports

Reverted back to the named export.

src/components/Settings/SettingDialog.tsx
denys-kolomiitsev: Reverted back to the named export as advised by comment of @mikitabut 
https://github.com/epam/ai-dial-chat/pull/346#discussion_r1427943085

src/components/Settings/ToggleFullWidth.tsx
denys-kolomiitsev: Changed to use only the min width to be alined with design.
image

denys-kolomiitsev: Thank you for advice.
Changed to the rounded-full

src/components/Settings/ToggleFullWidth.tsx
denys-kolomiitsev: Fixed width was needed to be alined with design
Changed to use flex values istead.

denys-kolomiitsev:
/deploy-review
>Deployment status: success

denys-kolomiitsev:
/deploy-review
>Deployment status: success

denys-kolomiitsev:
/deploy-review
>Deployment status: success

src/components/Settings/SettingDialog.tsx
```diff
@@ -1,22 +1,18 @@
...
 
+import { HighlightColor } from '@/src/types/common';
 import { Theme } from '@/src/types/settings';
 import { Translation } from '@/src/types/translation';
 
 import { useAppDispatch, useAppSelector } from '@/src/store/hooks';
 import { UIActions, UISelectors } from '@/src/store/ui/ui.reducers';
 
 import XMark from '../../../public/images/icons/xmark.svg';
+import ToggleFullWidth from '../Common/ToggleSwitch/ToggleSwitchLabeled';
```
mikitabut: Still think that default is not good to use, because now this file not about toggleFullWidth(now it's generic) but it still named such

denys-kolomiitsev: I think both approaches have advantages and disadvantages.

Changed to named export.

mikitabut (approved):

PR #356, opened 2023-12-17
chore: updated request api key for internal epam process

justrp:


src/components/Chat/RequestApiKeyDialog.tsx
```diff
@@ -375,7 +397,7 @@ export const RequestAPIKeyDialog: FC<Props> = ({ isOpen, onClose }) => {
               </span>
             </span>
             <span className="ml-1">
-              {t('Please provide name')}
+              {t('Please provide email')}
```
mikitabut, line 400: Please rename field below to have same understanding of what purpose of field below

Issue #293, opened 2023-12-05
It's impossible to upload file thru attachment functionality opened from chat panel if new chat or chat based on not DIAL RAG is opened


VolhaBazhkova:
1. Click on "+ New conversation"
2. Click on three dots at the bottom of the chat panel -> click on Attachment
3. Note that Supported types are not mentioned -> **bug**
![image](https://github.com/epam/ai-dial-chat/assets/111180458/d27e2105-a8f4-4a3e-b566-d597d4340ee0)
4. Click on Upload from device and select any file -> error appears -> **bug**
![image](https://github.com/epam/ai-dial-chat/assets/111180458/b2c75d2d-6f9c-4308-a805-d8d6706e8239)

**Note:** it works fine if DIAL RAG app is on the screen

**Expected result:** user should be able to upload files from attachment functionality on chat panel no matter what is opened on the screen (new chat or chat based on any other model)

VolhaBazhkova:
Verified on staging successfully. Please see also #366

PR #359, opened 2023-12-18
fix: it's impossible to upload file thru attachment functionality opened from chat panel if new chat or chat based on not dial rag is opened

Alexander-Kezik:
- Issue: https://github.com/epam/ai-dial-chat/issues/293


Alexander-Kezik:
/deploy-review
>Deployment status: success

mikitabut (approved):

Issue #347, opened 2023-12-15
Update files API integration


mikitabut:

VolhaBazhkova:
Verified on staging successfully

PR #364, opened 2023-12-18
feat: update files api integration

mikitabut:
#347


src/components/Files/PreUploadModal.tsx
```diff
@@ -100,7 +99,10 @@ export const PreUploadDialog = ({
     if (allowedTypes.includes('*/*')) {
       return [t('all')];
     }
-    return allowedTypes.map((mimeType) => extension(mimeType));
+    return allowedTypes
+      .map((mimeType) => getExtensionsList(mimeType))
+      .flat()
+      .map((type) => `.${type}`);
```
IlyaBondar: it's good candidate for util-method because the same block in `FileManagerModal.tsx`

src/pages/api/files/file/[...slug].ts
```diff
@@ -46,18 +46,23 @@ export const config = {
...
   token: JWT | null,
   res: NextApiResponse,
 ) {
   const readable = Readable.from(req);
-  const { path = '' } = req.query as { path: string };
-  const url = `${process.env.OPENAI_API_HOST}/v1/files${
-    path && `/${encodeURI(path)}`
-  }`;
+  const slugs =
+    typeof req.query.slug === 'string' ? [req.query.slug] : req.query.slug;
```
Comment (IlyaBondar) [inline src/pages/api/files/file/[...slug].ts]: maybe check that is array?
image

```suggestion
  const slugs =
    Array.isArray(req.query.slug) ? req.query.slug : [req.query.slug];
```

And it would be nice to create util-method

src/pages/api/files/file/[...slug].ts
```diff
@@ -83,10 +88,15 @@ async function handleGetRequest(
   token: JWT | null,
   res: NextApiResponse,
 ) {
-  const { path } = req.query as { path: string };
-  const url = `${process.env.OPENAI_API_HOST}/v1/files${
-    path && encodeURI(path)
-  }?purpose=content`;
+  const slugs =
+    typeof req.query.slug === 'string' ? [req.query.slug] : req.query.slug;
```
Comment (IlyaBondar) [inline src/pages/api/files/file/[...slug].ts]: the same (see previouis comment)

IlyaBondar (approved):

Issue #371, opened 2023-12-19
Shared and Published icons have unnecessary background if to navigate thru context menu; in Select conversation drop down list


VolhaBazhkova:
Need to fix the follwing issues:
1. If to open Select conversation drop down list and hover over the mouse like on the screenshot (below chat name) the background is incorrect
![image](https://github.com/epam/ai-dial-chat/assets/111180458/9b3012db-6a14-4eb4-82d3-70427d253321)

2. If to navigate thru chat context menu
![image](https://github.com/epam/ai-dial-chat/assets/111180458/54d16dcb-ee88-41e1-bd5c-01ecb6cd8c80)

VolhaBazhkova:
Verified on staging successfully

PR #372, opened 2023-12-19
fix: shared and published icons have unnecessary background if to navigate thru context menu in select conversation drop down list

Alexander-Kezik:
- Issue: https://github.com/epam/ai-dial-chat/issues/371

### DIFF
--- a/src/components/Chatbar/components/Conversation.tsx
+++ b/src/components/Chatbar/components/Conversation.tsx
@@ -44,9 +44,10 @@ import { v4 as uuidv4 } from 'uuid';
 interface ViewProps {
   conversation: Conversation;
   isHighlited: boolean;
+  isContextMenu: boolean;
 }
 
-export function ConversationView({ conversation, isHighlited }: ViewProps) {
+export function ConversationView({ conversation, isHighlited, isContextMenu }: ViewProps) {
   const modelsMap = useAppSelector(ModelsSelectors.selectModelsMap);
   const theme = useAppSelector(UISelectors.selectThemeState);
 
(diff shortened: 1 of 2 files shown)


Alexander-Kezik:
/deploy-review
>Deployment status: success

src/components/Chatbar/components/Conversation.tsx
```diff
@@ -399,6 +404,7 @@ export const ConversationComponent = ({ item: conversation, level }: Props) => {
           <ConversationView
             conversation={conversation}
             isHighlited={isHighlited}
+            isContextMenu={isContextMenu}
```
mikitabut: maybe it's better to have it inside isHighlighted property?

src/components/Common/Combobox.tsx
```diff
@@ -160,7 +160,7 @@ export const Combobox = ({
           (displayedItems?.length > 0 ? (
             displayedItems.map((item, index) => (
               <li
-                className={`flex cursor-pointer flex-col px-3 py-2 ${
+                className={`group flex cursor-pointer flex-col px-3 py-2 ${
```
mikitabut, line 163: not sure that adding group here is good solution, because every dropdown will have it now
Maybe it's better to add for item which passed to dropdown?

Alexander-Kezik: in this case group will work only when hovering on the model name (as it works now), not on the list item and the icon will lose background

IlyaBondar (approved):

src/components/Common/Combobox.tsx
IlyaBondar: @mikitabut , `ShareIcon` highligting doesn't work without `group-hover`

Issue #253, opened 2023-11-29
Limit attachments files and folders with special characters in name


mikitabut:
: ; , = /
On Chat UI - We should show error message when attachment has special characters in name 
On core API - reject such files with special characters in name (@avavilau to discuss with @Maxim-Gadalov)

VolhaBazhkova:
@IlyaBondar returning the ticket. 
We need to show error on "Upload from device" pop-up if user renames the files using ;:,/= symbols. 
Now we show error on Attach files pop-up and for the file with : Also file with / in name isn't shown at all. Files with ;,= is uploaded without any error

![image](https://github.com/epam/ai-dial-chat/assets/111180458/03761375-8235-4f7d-a250-ce742d553d1b)

VolhaBazhkova:
Verified on staging successfully

PR #373, opened 2023-12-19
fix: not allow incorrect symbols in file names (Issue #253)

IlyaBondar:
- Issue #253

### DIFF
--- a/src/components/Folder/Folder.tsx
+++ b/src/components/Folder/Folder.tsx
@@ -21,6 +21,7 @@ import classNames from 'classnames';
 
 import useOutsideAlerter from '@/src/hooks/useOutsideAlerter';
 
+import { excludeNotAllowedSymbols } from '@/src/utils/app/file';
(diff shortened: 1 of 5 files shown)


src/components/Folder/Folder.tsx
```diff
@@ -512,7 +513,9 @@ const Folder = <T extends Conversation | Prompt | DialFile>({
               className="mr-12 flex-1 overflow-hidden text-ellipsis bg-transparent text-left outline-none"
               type="text"
               value={renameValue}
-              onChange={(e) => setRenameValue(e.target.value)}
+              onChange={(e) =>
+                setRenameValue(excludeNotAllowedSymbols(e.target.value))
```
mikitabut, line 517: Maybe it's better to not replace invalid symbols but just reset new name and show error?

mikitabut: Discussed - will be changed in separate ticket

mikitabut (approved):

Issue #397, opened 2023-12-20
When having some incorrect mime types provided in list of inputAttachmentTypes you will see `.false`, but incorrect types should be filtered


mikitabut:
When having 
"inputAttachmentTypes": [
  "image/jpg",
  "image/jpeg",
  "image/png",
  "image/webp",
  "image/gif
]
next supported file extensions displayed `.false, .jpeg, .png, .webp, .gif`
We need to filter invalid mime types(image/jpg not exist)

VolhaBazhkova:
Closing the issue as it's technical dev update. Now there is no specific issues with attached files exist in DIAL, all types are available.

PR #401, opened 2023-12-20
fix: fix mime types with multiple extensions (Issue #397)

mikitabut:
#397

### DIFF
--- a/src/components/Files/FileManagerModal.tsx
+++ b/src/components/Files/FileManagerModal.tsx
@@ -354,7 +354,9 @@ export const FileManagerModal = ({
                   {t(
                     'Max file size up to 512 Mb. Supported types: {{allowedExtensions}}.',
                     {
-                      allowedExtensions: allowedExtensions.join(', '),
+                      allowedExtensions:
+                        allowedExtensions.join(', ') ||
+                        'no available extensions',
                     },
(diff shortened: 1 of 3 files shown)


src/components/Files/FileManagerModal.tsx
```diff
@@ -354,7 +354,9 @@ export const FileManagerModal = ({
                   {t(
                     'Max file size up to 512 Mb. Supported types: {{allowedExtensions}}.',
                     {
-                      allowedExtensions: allowedExtensions.join(', '),
+                      allowedExtensions:
+                        allowedExtensions.join(', ') ||
+                        'no available extensions',
```
IlyaBondar, line 359: Maybe should be ` all` by default?

IlyaBondar (approved):

denys-kolomiitsev (approved):

IlyaBondar:
are all extensions now available for `image/jpeg`?
image

src/components/Files/FileManagerModal.tsx
mikitabut: For all types we have special `*/*` case, so this is not real case, only when incorrect configuration

mikitabut:
> 

Yes

Issue #129, opened 2023-11-14
Need to add addon icon to the received stages


VolhaBazhkova:
> Addon's stages for assistants with Addon icon and name

Example (the link is available for DIAL dev team only): [DIAL Chat – Figma](https://www.figma.com/file/lwL5jneveZczAj5RLcLFH2/DIAL-Chat?type=design&node-id=1483-7257&mode=design&t=GmNhFL5pJvFGzgS3-0)

From Figma:
![image](https://github.com/epam/ai-dial-chat/assets/111180458/cabc2a17-289c-4b42-ab79-0d8ab62cc451)

VolhaBazhkova:
Verified on staging server successfully. Addon name and tooltip on hover exist on stage.
![image](https://github.com/epam/ai-dial-chat/assets/111180458/e0716549-1187-433c-b53c-1089831d2194)
![image](https://github.com/epam/ai-dial-chat/assets/111180458/f466f9ac-758b-4650-832e-2b873fc756f0)

AnastasiiaHarkot:
**QA Verification — Fixed ✅**

**Verified by:** AnastasiiaHarkot
**Date:** 2026-09-16
**Result:** Addon icon is now properly displayed on the received stages.

**Verdict: Fixed** — verified.

PR #481, opened 2024-01-09
feat: add icon to the received stages (Issue #129)

Alexander-Kezik:
- Issue: https://github.com/epam/ai-dial-chat/issues/129


Alexander-Kezik:
/deploy-review
>Deployment status: success

src/components/Chat/MessageStage.tsx
```diff
@@ -17,6 +23,17 @@ interface StageTitleProps {
 }
 
 const StageTitle = ({ isOpened, stage }: StageTitleProps) => {
+  const [addon, setAddon] = useState<OpenAIEntityAddon | undefined>();
+  const addonsMap = useAppSelector(AddonsSelectors.selectAddonsMap);
+
+  const match = stage.name?.match(/^[^(]*/);
```
IlyaBondar, line 29: this regex match everything except including `(`
even empty string
e.g.
match(`123(45`) -> `123`
match(`(12345`) -> ``
Is it expected?

Alexander-Kezik: We get a stage from core, whose name for the addon will look like this: addonName(question). If it finds a match with the addonMap, then we will show the icon. For now, only after receiving the part before "(" we will be able to show the icon.

IlyaBondar (approved):

Issue #458, opened 2024-01-04
Add missing gradient for error message


IlyaBondar:
#### Actual result:
![image](https://github.com/epam/ai-dial-chat/assets/20163971/77f2af39-c318-4d11-b84b-869687cd8c0b)
#### Expected result:
![image](https://github.com/epam/ai-dial-chat/assets/20163971/52403dc7-7703-44bf-b8b9-e67bd88dd107)

To reproduce you can import such file:
[chatbot_ui_conversation_1-4.json](https://github.com/epam/ai-dial-chat/files/13830848/chatbot_ui_conversation_1-4.json)

VolhaBazhkova:
Return to @IlyaBondar as we need to fix one more case:
![image](https://github.com/epam/ai-dial-chat/assets/111180458/4cdcfaff-7479-4d9a-89a7-00d4bfb565fe)

VolhaBazhkova:
Now the fix is perfect with "Not allowed model" error on staging env.
@IlyaBondar Can you check one more time the fix with text as there is still a little bit of text shown:
![image](https://github.com/epam/ai-dial-chat/assets/111180458/68acbe57-f0a8-4083-b738-d950b38c1a39)

VolhaBazhkova:
Verified on staging server successfully.

PR #502, opened 2024-01-11
fix: fix gradient over text (Issue #458)

IlyaBondar:
- Issue #458 
image
New: `via-40%` (see [here](https://tailwindcss.com/docs/gradient-color-stops#specifying-stop-positions))

### DIFF
--- a/src/styles/globals.css
+++ b/src/styles/globals.css
@@ -123,7 +123,11 @@ pre:has(div.codeblock) {
   }
 
   .gradient {
-    @apply absolute bottom-0 left-0 w-full border-transparent bg-gradient-to-b from-transparent via-layer-1 via-40% to-layer-1 pt-6 md:pt-2;
+    @apply border-transparent bg-gradient-to-b from-transparent via-layer-1 via-40% to-layer-1;
+  }
+
+  .gradient.gradient-bottom {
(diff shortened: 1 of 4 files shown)


IlyaBondar:
/deploy-review
>Deployment status: success

src/styles/globals.css
```diff
@@ -121,6 +121,10 @@ pre:has(div.codeblock) {
   .checkbox {
     @apply relative m-0 mr-2 inline h-4 w-4 shrink-0 appearance-none rounded-sm border border-primary text-accent-primary checked:border-accent-primary hover:border-accent-primary focus-visible:outline-none;
   }
+
+  .gradient {
+    @apply absolute bottom-0 left-0 w-full border-transparent bg-gradient-to-b from-transparent via-layer-1 via-40% to-layer-1 pt-6 md:pt-2;
```
mikitabut: I'm not sure that you need to have all classes inside gradient class - I think that having only related to gradient classes (`bg-gradient-to-b from-transparent via-layer-1 via-40% to-layer-1`) should be enough and will be more easy to use

Also it better to name it like top-bottom-gradient because we also have left-right gradient already and we can have more in future

mikitabut (approved):

PR #511, opened 2024-01-12
fix: check model id passed by user

mikitabut:
User have been able to pass some modelId which could affect final URL for rating. Now we are checking is this modelId exists and only after that we will call rate message endpoint


denys-kolomiitsev (approved):

IlyaBondar (approved):

IlyaBondar (changes requested):
Please, add Issue in PR title and description

IlyaBondar (approved):

Issue #512, opened 2024-01-12
After successful login in Overlay "Not allowed models" error message is shown for a moment. The same if to refresh browser when Dial chat is opened.


VolhaBazhkova:
1. Open DIAl app with any opened chat or just create new chat
2. Click on Refresh browser -> "Not allowed model" error appears for a moment -> **bug** 

**Expected result:** should not "Not allowed model" error appear 

Please see attached video (started from 39 sec) - it's issue example with Overlay

https://github.com/epam/ai-dial-chat/assets/111180458/54ad332b-8da4-4623-a829-52c23ba2493f

VolhaBazhkova:
Verified on staging successfully (no error on browser refresh and after overlay login)

Issue #100, opened 2023-11-13
Overlay. Changing model setting for overlay require removing local storage for applying it


VolhaBazhkova:
Changing model setting for overlay require removing local storage for applying it.  
![image](https://github.com/epam/ai-dial-chat/assets/111180458/957261a7-bfb9-4b02-b619-7e749f2bbb3c)

**Expected:** model should applied correctly without local storage clearing

[01/12/2024] After the fix of #425 I see the following picture
![image](https://github.com/epam/ai-dial-chat/assets/111180458/2e7a68f9-eef9-413b-bb34-a23362af160e)

Overlay works in incognito mode only. User has to clear local storage, which we try to avoid.

VolhaBazhkova:
Verified successfully using the following steps:
1. Open site with overlay connected
2. Open browser console -> local storage -> conversationHistory 
3. Update to not existing modelID -> Enter
4. Refresh browser
5. Check that conversationHistory contains correct modelID, but one set in 3rd step

PR #530, opened 2024-01-15
fix: overlay infinite loader and not allowed model (Issue #100, #512)

mikitabut:
- Issue #100 
- Issue #512


mikitabut:
/deploy-review
>Deployment status: success

src/store/overlay/overlay.epics.ts
```diff
@@ -359,9 +370,18 @@ const notifyHostGPTMessageStatus: AppEpic = (_, state$) =>
...
-const notifyHostAboutReadyEpic: AppEpic = (action$) =>
-  action$.pipe(
-    filter(ModelsActions.getModelsSuccess.match),
+const notifyHostAboutReadyEpic: AppEpic = (_action$, state$) =>
+  state$.pipe(
+    filter((state) => {
+      const isAuthDisabled = SettingsSelectors.selectIsAuthDisabled(state);
+      const isShouldLogin = AuthSelectors.selectIsShouldLogin(state);
+      const authStatus = AuthSelectors.selectStatus(state);
+      if (isAuthDisabled || (authStatus !== 'loading' && isShouldLogin)) {
```
IlyaBondar, line 379: Does it make sense to create constants for `loading`?

IlyaBondar (approved):

src/store/overlay/overlay.epics.ts
mikitabut: Not sure, status field is more flexible in such cases

Issue #585, opened 2024-01-22
"Move to" contains folders from "Shared with me" and "Organization sections"


VolhaBazhkova:
1. Prepare Folders with chats and other nested folders/chat
2. Share and publish the root folder to have Shared with me and Organization sections
3. Create a new chat to have it Today section
4. Hover the chat > click on three dots > hover "Move to" item in the menu
5. Note that "Move to" contains all the folders from Shared with me and Organization sections -> **bug**

**Expected result:** Move to is available for my chats only and are allowed to be moved thu my folders (folders, created by me). Move to should NOT contain folders from Shared with me and Organization sections 

Note: Remember to fix and check Prompts

VolhaBazhkova:
Verified on staging server successfully
![image](https://github.com/epam/ai-dial-chat/assets/111180458/b84a3962-7c2c-49b8-9a7e-af3279034a6a)

![image](https://github.com/epam/ai-dial-chat/assets/111180458/d12ad7a2-f0db-4040-9900-2263679da63c)

PR #587, opened 2024-01-23
fix: show in MoveTo only own folders (Issue #585)

IlyaBondar:
- Issue #585

### DIFF
--- a/src/components/Chatbar/components/Conversation.tsx
+++ b/src/components/Chatbar/components/Conversation.tsx
@@ -14,7 +14,7 @@ import {
 import classNames from 'classnames';
 
 import { hasParentWithFloatingOverlay } from '@/src/utils/app/modals';
-import { getMyItemsFilters } from '@/src/utils/app/search';
+import { defaultMyItemsFilters } from '@/src/utils/app/search';
 
 import { Conversation } from '@/src/types/chat';
 import { FeatureType } from '@/src/types/common';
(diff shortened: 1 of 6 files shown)


src/components/Chatbar/components/Conversation.tsx
```diff
@@ -108,7 +109,14 @@ export const ConversationComponent = ({ item: conversation, level }: Props) => {
   const isReplay = useAppSelector(
     ConversationsSelectors.selectIsReplaySelectedConversations,
   );
-  const folders = useAppSelector(ConversationsSelectors.selectFolders);
+  const folders = useAppSelector((state) =>
+    ConversationsSelectors.selectFilteredFolders(
+      state,
+      getMyItemsFilters(),
+      '',
+      true,
```
mikitabut, line 117: Its not good to pass booleans as arguments
Is it possible to change just arguments to object with this arguments inside?

src/utils/app/folders.ts
```diff
@@ -204,14 +204,25 @@ export const getPathToFolderById = (
 };
 
 export const getFilteredFolders = (
-  folders: FolderInterface[],
+  allFolders: FolderInterface[],
```
mikitabut: I understand that this is not your changes, but could you please also change this 6 params to object - it will be more easier to pass arguments inside and you will not have issues with order and it will be more visible what params passed

denys-kolomiitsev (dismissed):

src/components/Chatbar/components/Conversation.tsx
IlyaBondar: if we pack argument into object memorization won't work, because it always will be new object/link

IlyaBondar:
/deploy-review
>Deployment status: success

src/components/Chatbar/components/Conversation.tsx
mikitabut: We need to avoid such things, but let's go with what we have for now

mikitabut (approved):

Issue #584, opened 2024-01-22
Need to forbid using drag&drop functionality inside "Shared with me", "Organization" and from/to these sections


VolhaBazhkova:
1. Prepare Folders with chats and other nested folders/chat
2. Share and publish the root folder to have Shared with me and Organization sections
3. Using drag & drop change the structure in Shared with me and Organization sections -> it's allowed -> **issue1**
4. Using drag & drop move folders from Shared with me and Organization sections to Pinned/or other sections ->  it's allowed -> **issue2**
5. Using drag & drop move folders/chats from Pinned to Shared with me and Organization sections ->  it's allowed -> **issue3**

**Expected result:** Need to forbid using drag&drop functionality inside "Shared with me", "Organization" and from these sections

Note: Remember to fix and check Prompts

VolhaBazhkova:
Verified on staging server successfully

PR #596, opened 2024-01-24
feat: set external entities as not draggable (Issue #584)

IlyaBondar:
- Issue #584

### DIFF
--- a/src/utils/app/share.ts
+++ b/src/utils/app/share.ts
@@ -1,4 +1,4 @@
-import { FeatureType, ShareEntity } from '@/src/types/common';
+import { Entity, FeatureType, ShareEntity } from '@/src/types/common';
 import { SharingType } from '@/src/types/share';
 
 import {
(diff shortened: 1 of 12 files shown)


IlyaBondar:
/deploy-review
>Deployment status: success

src/utils/app/share.ts
```diff
@@ -57,7 +57,7 @@ export const getUnpublishActionByType = (type: SharingType) => {
   }
 };
 
-export const isEntityExternal = (entity: ShareEntity) =>
+export const isThisEntityExternal = (entity: ShareEntity) =>
```
mikitabut: What is the difference between isThisEntityExternal and isExternalEntity? 
It's not clear when to use them
Is it possible to rename somehow?

mikitabut: Maybe it's possible to use full check for external everywhere?

IlyaBondar: @mikitabut , I specially added `This` word to have difference when we check entity itself only (and when we check entity and parents)

IlyaBondar:
/deploy-review
>Deployment status: success

IlyaBondar:
/deploy-review
>Deployment status: success

mikitabut (approved):

Issue #604, opened 2024-01-25
If to drag and drop Folder_parent into Folder_child both folders disappears. Applicable for chat and prompt folders.


VolhaBazhkova:
If to drag and drop Folder_parent into Folder_child both folders disappears. Applicable for chat and prompt folders.

1. Create a structure:
- Folder_parent 
--Folder_child

2. Drag Folder_parent and drop it in Folder_child -> both folders disappears from UI -> **bug**

Expected result: not allow to move parent folder in child folder

VolhaBazhkova:
Verified on staging server successfully

PR #606, opened 2024-01-25
fix: not allow to move parent folder in child folder (Issue #604)

IlyaBondar:
- Issue #604


src/components/Folder/Folder.tsx
```diff
@@ -262,12 +262,24 @@ const Folder = <T extends Conversation | Prompt | DialFile>({
           getFolderMoveType(featureType),
         );
 
+        const draggedFolder = JSON.parse(folderData);
```
mikitabut: Looks like folder data can be undefined if I dragging conversation over folder

Move it under condition below

mikitabut (changes requested):

Alexander-Kezik (dismissed):

mikitabut (dismissed):

Alexander-Kezik (approved):

PR #620, opened 2024-01-29
fix: added curation for union NPM package

adubovik:
As per [curations doc](https://github.com/oss-review-toolkit/ort/blob/61fbc328ae18392c0f2484d2f09e18454cefc777/website/docs/configuration/ort-yml.md#curations)

### DIFF
--- a/.github/workflows/pr_check_tests.yml
+++ b/.github/workflows/pr_check_tests.yml
@@ -4,6 +4,9 @@ on:
   pull_request:
     branches: [ development, release-* ]
 
+env:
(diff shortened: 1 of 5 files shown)


adubovik:
Can't integrate curations into `.ort.yml` since it's prohibited:

```
WARNING: PACKAGE_CURATION_IN_ORT_YML - The use of package curations is not allowed in the *.ort.yml file.
```

Alternatively, when the curations are declared in `config.yaml`, ORT doesn't pick up the file, because it isn't configured to look at this conf.

PR #646, opened 2024-02-05
feat(chat-e2e): introduced unique users for e2e worker

irinakartun:


apps/chat-e2e/config/playwright.config.ts
```diff
@@ -11,16 +9,19 @@ export default defineConfig({
   /* Run tests in files in parallel */
   fullyParallel: false,
   /* Opt out of parallel tests on CI. */
-  workers: 6,
+  workers: process.env.E2E_HOST ? 6 : 2,
```
mikitabut: What do you think about adding this as env variable?

mikitabut (dismissed):

apps/chat-e2e/config/playwright.config.ts
irinakartun: I'm not sure that's a good idea because every time there will be a need to adjust number of workers, we will need to update CI as well

mikitabut (approved):

irinakartun:
/deploy-review
>Environment URL: https://chat-ai-dial-chat-pr-646.nightly-test.deltixhub.io
>E2E tests status: success

Issue #592, opened 2024-01-23
[Attachments] Need to forbid # symbol


VolhaBazhkova:
It is impossible to load file with # symbol in the name. The following screen is shown (nothing happens). no error in console

![image](https://github.com/epam/ai-dial-chat/assets/111180458/7576235b-2628-47a5-a922-9ed8e6d7858d)

Expected: need to show error if to load file with # in the name
Note: similar fixes were here #253 #417

Alexander-Kezik:
need to forbid \ symbol also

VolhaBazhkova:
Now on staging the symbols are forbidden. But the change is going to be reverted

PR #647, opened 2024-02-05
fix: forbid # and \ symbols

Alexander-Kezik:
- Issue: https://github.com/epam/ai-dial-chat/issues/592

### DIFF
--- a/apps/chat/src/utils/app/file.ts
+++ b/apps/chat/src/utils/app/file.ts
@@ -99,7 +99,7 @@ export const getFilesWithInvalidFileType = (
     ? []
     : files.filter((file) => !isAllowedMimeType(allowedFileTypes, file.type));
 };
-export const notAllowedSymbols = ':;,=/';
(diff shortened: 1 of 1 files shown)


apps/chat/src/utils/app/file.ts
```diff
@@ -99,7 +99,7 @@ export const getFilesWithInvalidFileType = (
     ? []
     : files.filter((file) => !isAllowedMimeType(allowedFileTypes, file.type));
 };
-export const notAllowedSymbols = ':;,=/';
+export const notAllowedSymbols = ':;,=/#\\\\';
```
mikitabut, line 102: Why 4 \ ?

mikitabut (approved):

apps/chat/src/utils/app/file.ts
Alexander-Kezik: 4 times because \\ for escaping, and 2 more for escaping in a regular expression, in which it is then used.

mikitabut (approved):

Issue #640, opened 2024-02-02
Need to integrate export/import with Stateful API


IlyaBondar:
Need to implement:
- [x] export single chat: upload all entities for export from API (chats, folders, attachments)
- [x] export history: upload all entities for export from API (chats, prompts, folders)
- [x] export: need to upgrade export version (`"version": 5`) and replace "chatbot_ui" to "ai_dial_chat"(see #594 )
- [x] import: need to generate id from GUID-format to path-format(see `getPathToFolderById` + `generateConversationId`/`generatePromptId`) on the fly for export format less version 5 (see #633)
- [x] import: save imported data in API

Additionally:
- [x] see #594 
- [ ] see #609
- [x] see #651 
- [x] see #652

VolhaBazhkova:
Verified on staging. Import/export works fine without attachments.
See also #797 
See #819 for export/import with attachments

PR #692, opened 2024-02-13
fix: add empty folders to the export(issue#640)

denys-kolomiitsev:
fixed export of empty folders and export with attachments

- #640

### DIFF
--- a/apps/chat/src/store/import-export/importExport.epics.ts
+++ b/apps/chat/src/store/import-export/importExport.epics.ts
@@ -19,7 +19,7 @@ import {
 
 import { combineEpics } from 'redux-observable';
 
-import { filterOnlyMyEntities } from '@/src/utils/app/common';
+import { combineEntities, filterOnlyMyEntities } from '@/src/utils/app/common';
 import { BucketService } from '@/src/utils/app/data/bucket-service';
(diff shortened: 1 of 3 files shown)


apps/chat/src/store/import-export/importExport.epics.ts
```diff
@@ -143,13 +143,20 @@ const exportConversationsEpic: AppEpic = (action$, state$) =>
...
         Array.from(
           new Set(foldersIds.flatMap((id) => getAllPathsFromPath(id))),
         ),
         FolderType.Chat,
       );
 
+      const allFolders = ConversationsSelectors.selectFolders(state$.value);
+      const emptyFolders = allFolders.filter(
+        ({ id }) => !foldersIds.some((folderId) => folderId === id),
+      );
```
IlyaBondar: this can be removed - `combineEntities` will merge by ID (ignore second duplicate)
just: 
```javascript
const folders = combineEntities(foldersWithConversation, allFolders);
```

denys-kolomiitsev: Thank you! 
Changed.

mikitabut (approved):

Issue #772, opened 2024-02-16
[Integration] all the attachments are deleted from message edit mode if to click on 'x' for one file only


VolhaBazhkova:
https://github.com/epam/ai-dial-chat/assets/111180458/e5b03ba8-eb05-453b-ac59-f43b6e42d2af

VolhaBazhkova:
Verified on staging successfully

PR #811, opened 2024-02-20
fix(chat): fix files deleting while edit message (Issue #772)

Alexander-Kezik:
- Issue: https://github.com/epam/ai-dial-chat/issues/772

### DIFF
--- a/apps/chat/src/utils/app/file.ts
+++ b/apps/chat/src/utils/app/file.ts
@@ -152,7 +152,7 @@ export const getDialFilesFromAttachments = (
       const { absolutePath, name } = parseAttachmentUrl(attachment.url);
 
       return {
-        id: constructPath(absolutePath, name),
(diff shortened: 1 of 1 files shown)


mikitabut (dismissed):

apps/chat/src/utils/app/file.ts
```diff
@@ -152,7 +152,7 @@ export const getDialFilesFromAttachments = (
       const { absolutePath, name } = parseAttachmentUrl(attachment.url);
 
       return {
-        id: absolutePath,
+        id: constructPath(absolutePath, name),
```
IlyaBondar: Just use `attachment.url` instead because it's absolutely equals  to `constructPath(absolutePath, name)`

IlyaBondar (changes requested):

IlyaBondar (approved):

Issue #833, opened 2024-02-22
[Core][Share] Folder from "Shared with me" doesn't disappear automatically if to rename original one. Note that chat located in shared folder disappears in this case


VolhaBazhkova:
1. By user1 create folder with chat inside -> share the folder
2. By user2 click on the link -> folder with char appears in Shared with me
3. By user1 renamed the folder -> shared icon disappears (ok)
4. By user2 refresh the browser and note that the chat disappears (ok), but folder is still in "Shared with me" structure -> **bug**

**Expected result:** folder with its elements should disappear from "Shared with me" if the original shared folder was renamed

Note: user2 can delete such folder from Shared with me thru context menu

IlyaBondar:
Related to https://github.com/epam/ai-dial-core/issues/221

VolhaBazhkova:
Verified on staging successfully

PR #893, opened 2024-03-01
fix(chat): revoke access to folder if folder renamed (Issue #833)

mikitabut:
**Description:**

Revoke access to folder if rename or delete

Issues:

- Issue #833

**Checklist:**

- [x] the pull request name ends with `(Issue #<TICKET_ID>)` (comma-separated list of issues)

### DIFF
--- a/apps/chat/src/store/share/share.epics.ts
+++ b/apps/chat/src/store/share/share.epics.ts
@@ -753,6 +753,92 @@ const discardSharedWithMeFailEpic: AppEpic = (action$) =>
     }),
   );
 
+const deleteSharedConversationFolderEpic: AppEpic = (action$, state$) =>
+  action$.pipe(
+    filter(ConversationsActions.deleteFolder.match),
+    switchMap(({ payload }) => {
+      const folders = ConversationsSelectors.selectFolders(state$.value);
+      const isSharedFolder = folders.find(
+        (folder) => folder.id === payload.folderId,
+      )?.isShared;
+
+      return payload.folderId && isSharedFolder
+        ? of(
+            ShareActions.revokeAccess({
+              resourceId: payload.folderId,
+              featureType: FeatureType.Chat,
+              isFolder: true,
+            }),
+          )
+        : EMPTY;
+    }),
+  );
+
+const deleteSharedPromptFolderEpic: AppEpic = (action$, state$) =>
+  action$.pipe(
+    filter(PromptsActions.deleteFolder.match),
+    switchMap(({ payload }) => {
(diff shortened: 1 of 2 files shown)


mikitabut:
/deploy-review
>Environment URL: https://chat-ai-dial-chat-pr-893.nightly-test.deltixhub.io
>E2E tests status: success

apps/chat/src/store/share/share.epics.ts
```diff
@@ -753,6 +753,92 @@ const discardSharedWithMeFailEpic: AppEpic = (action$) =>
     }),
   );
 
+const deleteSharedConversationFolderEpic: AppEpic = (action$, state$) =>
```
Alexander-Kezik: Is it possible to combine epics and pass feature type and folders as payload?

apps/chat/src/store/share/share.epics.ts
```diff
@@ -780,4 +866,9 @@ export const ShareEpics = combineEpics(
 
   triggerGettingSharedListingsConversationsEpic,
   triggerGettingSharedListingsPromptsEpic,
+
+  deleteSharedConversationFolderEpic,
+  deleteSharedPromptFolderEpic,
+  renameSharedConversationFolderEpic,
+  renameSharedPromptFolderEpic,
```
IlyaBondar: what's difference between `delete` and `rename`? Do we really need pair epics for these cases?
If `key` became different then execute `revoke`

apps/chat/src/store/share/share.epics.ts
mikitabut: Not sure that combining here different epics is good idea, because I will need to to do some ugly logic to diff them, but now its simple and easy to understand

apps/chat/src/store/share/share.epics.ts
mikitabut: the difference is payload, actions on what they triggered and conditions when we need to call revoke, that's why we need separate actions

apps/chat/src/store/share/share.epics.ts
IlyaBondar: yes, it looks like it can be combined in one epic with `featureType`

IlyaBondar: @mikitabut , but in current case we need to maintain 4 similar epics which do the same (difference in 1-2 lines)

mikitabut: Fixed

denys-kolomiitsev (approved):

PR #902, opened 2024-03-04
feat(chat): Correct overlay systemPrompt setting just before sendMessage

4llower:
systemPrompt in overlay we're using to share host application state to dial applications.

Why it's improving current behavior:

1. It's not using updateConversation, which DIAL uses in 40+ places. Removes all bugs that related state dissapear of some actions. We 100% sure that application state will pass to /chat request.
2. It's not clearing systemPrompt if we change conversations or clear conversation. 
3. Full backward compatibility


apps/chat/src/store/conversations/conversations.epics.ts
```diff
@@ -1113,76 +1114,109 @@ const sendMessageEpic: AppEpic = (action$, state$) =>
...
-            conv.id !== payload.conversation.id,
-        ),
-        Math.max(selectedConversationIds.indexOf(payload.conversation.id), 0),
-        true,
-      );
+        /*
+          Overlay needs to share host application state information
+          We storing state information in systemPrompt (message with role: Role.System)
+        */
+        if (overlaySystemPrompt) {
```
mikitabut: please add addditional check is it overlay or not just to be sure that it will not called in normal mode

Moreover please extract this inner implementation to some method outside of this epic file

apps/chat/src/store/conversations/conversations.epics.ts
```diff
@@ -1113,76 +1114,109 @@ const sendMessageEpic: AppEpic = (action$, state$) =>
...
+          );
 
-      return {
-        oldConversationId: payload.conversation.id,
-        updatedConversation,
-        modelsMap,
-        assistantMessage,
-      };
-    }),
+          currentMessages.push(overlaySystemPromptMessage);
```
mikitabut: Maybe it's better to replace previous system message and place new on same index?
Or if it was not presented place on zero place in this array?

apps/chat/src/store/conversations/conversations.epics.ts
4llower: Fixed.

apps/chat/src/store/conversations/conversations.epics.ts
4llower: Fixed

mikitabut (dismissed):

mikitabut (approved):

PR #1644, opened 2024-06-24
fix(ovelay): fix error in terminal (Issue#1651)

denys-kolomiitsev:
**Description:**

- Fix error in terminal console "window not defined".

Issues:

- Issue #1651  

**Checklist:**

- [x] the pull request name ends with `(Issue #<TICKET_ID>)` (comma-separated list of issues)


libs/overlay/src/lib/ChatOverlay.ts
```diff
@@ -1,3 +1,5 @@
+'use client';
```
mikitabut: Is this really working? This is react specific thing or I'm wrong?

libs/overlay/src/lib/ChatOverlay.ts
```diff
@@ -358,7 +360,7 @@ export class ChatOverlay {
    */
   destroy() {
     window.removeEventListener('message', this.process);
-    this.iframeInteraction.fail('Chat Overlay destroyed');
+    this.iframeInteraction.complete();
```
mikitabut: Not sure that this is good idea. Actually, your iframe is not complete at this moment, but failing. Maybe we should just not log this as error, but still fail promise?
Moreover if this iframe interaction completed that means that user can send chat overlay options but this is not right scenario - they should do this only when it's really complete

libs/overlay/src/lib/ChatOverlay.ts
IlyaBondar: @mikitabut , it's NextJS feature

apps/chat/src/utils/app/overlay.ts
```diff
@@ -21,6 +21,7 @@ export function sendPMResponse(
   type: OverlayRequests,
   requestParams: PostMessageRequestParams,
 ) {
+  if (typeof window === 'undefined') return;
```
IlyaBondar, line 24: can we move all `window`-related code in separate file and mark by `use client`?

libs/overlay/src/lib/ChatOverlay.ts
mikitabut: This ChatOverlay.ts can be loaded outside of NextJS, so we shouldn't use it here

mikitabut: @IlyaBondar looks like its react :)
https://react.dev/reference/rsc/use-client

denys-kolomiitsev: [This is Next JS feature](https://nextjs.org/docs/app/building-your-application/rendering/client-components#using-client-components-in-nextjs)

apps/chat/src/utils/app/overlay.ts
denys-kolomiitsev: You a right, it's NextJS feature.
However, it doesn't work in this case.
For now only this solution works.

libs/overlay/src/lib/ChatOverlay.ts
denys-kolomiitsev: You are right. Reverted.

IlyaBondar (dismissed):

mikitabut (approved):

denys-kolomiitsev:
/deploy-review
>Environment URL: https://chat-ai-dial-chat-pr-1644.nightly-test.deltixhub.io
>E2E tests status: success

PR #6888, opened 2026-05-26
fix(overlay)!: prevent stale token 401s after auth refresh (Issue #6887)

### LINKED ISSUE #6887: Overlay mode sends stale bearer token after refresh and returns Core 401 on first post-expiry request
### DIAL chat version

0.45.1

### How to reproduce

1. Configure DIAL Chat to use the Auth0 provider with a refresh token available and `offline_access` included in the requested scope.
2. Run DIAL Chat in overlay mode with `IS_IFRAME=true`.
3. Sign in and wait for the Auth0 access token to expire. In the reported environment this happened after about 15 minutes.
4. After expiry, perform the first authenticated action, such as sending a chat message.
5. Observe the network and server behavior for that first post-expiry request.
Additional notes from the reproduced flow:
1. `getServerSession()` can refresh the token and set a new cookie on the response.
2. The same request path can still read the old token from the incoming request cookie.
3. In overlay mode, server validation may allow the request to continue even when the session contains `RefreshAccessTokenError`.
4. The downstream DIAL Core call then receives an expired bearer token.

### Actual result

The first request after token expiry can fail even though a refresh token is available.

Observed behavior:
- The request refreshes auth state on the server, but the outbound bearer token can still come from the stale request cookie.
- DIAL Core returns `401` for that first post-expiry request.
- In overlay mode, the user can be sent back to the login screen and then shortly after signed in again automatically because the upstream IdP session is still active.
- If a refresh attempt fails, the in-memory refresh lock can remain stuck in a refreshing state, causing later requests for the same user to wait and eventually time out instead of retrying cleanly.
In short, the system can both forward a stale token on the first request after expiry and leave refresh coordination in a bad state after a failed refresh.

### Expected result

The first request after token expiry should succeed when a valid refresh token is available.

Expected behavior:
- The same request that triggers server-side refresh should use the refreshed bearer token instead of the stale token from the incoming cookie.
- If token refresh fails, the refresh lock should be cleared so the next request can retry normally.
- Overlay API routes should fail early on invalid auth inside DIAL Chat instead of proxying an expired token to DIAL Core and surfacing a downstream `401`.
- The user should not see the first post-expiry action fail due to this auth race.

### Confidential information

- [x] I confirm that do not share any confidential information

blepam:
## Description of changes

This PR fixes the overlay auth expiry flow when OAuth access tokens expire and a refresh token is available.

The main issue was that a request could refresh the session through `getServerSession()` and still forward the stale bearer token from the incoming request cookie on that same request. In overlay mode, that led to downstream DIAL Core `401` responses on the first action after expiry.

This PR makes three related fixes:

- `getFullToken()` now checks the in-process refresh state and prefers the refreshed token when one was produced during the current server auth flow.
- The refresh lock lifecycle in `auth-callbacks.ts` and `nextauth-client.ts` is hardened so failed refresh attempts clear the lock, while timeout waiters do not clear a lock they do not own.
- `validateServerSession()` now enforces strict auth validation for API routes, including overlay mode, so invalid sessions fail early in DIAL Chat instead of proxying expired tokens to DIAL Core.

This PR also adds focused unit coverage for:

- refresh lock storage and cleanup
- overlay session validation behavior
- refresh-aware token resolution in `server.ts`
- refresh success and failure paths in `auth-callbacks.ts`

## Applicable issues

- fixes #6887

## UI changes

None.

## Breaking change (overlay mode only, low impact)

Overlay API routes (`IS_IFRAME=true`) now reject invalid/expired sessions with `401` at the Chat layer instead of proxying them to DIAL Core. Those requests were already failing downstream. Standard overlay usage is unaffected and should behave better after OAuth token refresh.

## Checklist

- [x] the pull request name ends with `(Issue #<ISSUE_ID>)` (comma-separated list of issues)


denys-kolomiitsev:
This is a breaking behavior change for all overlay API routes. Any overlay integration relying on the previous lenient behavior will now get hard 401s on expired sessions. This is likely the correct fix, but it should be called out explicitly in the PR description.

blepam:
@denys-kolomiitsev truth be told, I can't really think of any situation where the previous lenient behavior would be desired. 

The previous code didn't give overlay integrations a supported way to keep working. Requests went to DIAL Core with an expired, stale, or empty bearer token and failed with 401. The new stricter validation just fails earlier and more clearly.

Regardless, I've called out the breaking change in the description.

apps/chat/src/utils/auth/nextauth-client.ts
```diff
@@ -65,13 +65,31 @@ class NextClient {
...
   ): void {
+    globalObj._refreshTokenMap = globalObj._refreshTokenMap || {};
     globalObj._refreshTokenMap[userId] = refreshToken;
   }
 
+  /**
+   * Clears the refresh lock for a user after a failed refresh attempt,
+   * preserving the last known token so subsequent waiters can retry.
+   */
+  public static clearRefreshToken(userId: string): void {
```
denys-kolomiitsev: I think `clearRefreshToken` is misleading method name.

The method does not clear the entry — it sets `isRefreshing: false` while preserving the token. A caller reading only the name would expect the entry to be removed. A name like `releaseRefreshLock` or `resetRefreshingState` better describes the semantics.

blepam: Agreed - nice catch. I've updated the name.

denys-kolomiitsev:
Please run `npm run format:fix` to fix format

denys-kolomiitsev:
/deploy-review

>GitHub actions run: [26877445546](https://github.com/epam/ai-dial-ci/actions/runs/26877445546)

| [deploy-review](https://chat-ai-dial-chat-pr-6888.gke.test.dial.parts) | [Success ✅](https://github.com/epam/ai-dial-ci/actions/runs/26877445546/job/79268405008) |
| overlay | [Success ✅](https://github.com/epam/ai-dial-ci/actions/runs/26877445546/job/79273001323) |
| chat | [Success ✅](https://github.com/epam/ai-dial-ci/actions/runs/26877445546/job/79273001483) |

denys-kolomiitsev (approved):

denys-kolomiitsev:
/deploy-review

>GitHub actions run: [26943068832](https://github.com/epam/ai-dial-ci/actions/runs/26943068832)

| [deploy-review](https://chat-ai-dial-chat-pr-6888.gke.test.dial.parts) | [Success ✅](https://github.com/epam/ai-dial-ci/actions/runs/26943068832/job/79489099690) |
| chat | [Success ✅](https://github.com/epam/ai-dial-ci/actions/runs/26943068832/job/79493874345) |
| overlay | [Success ✅](https://github.com/epam/ai-dial-ci/actions/runs/26943068832/job/79493874450) |

PR #7407, opened 2026-06-24
feat: add file manager tabs and shared/public file listing endpoints

valerydluski:
## Applicable issues

- fixes #<ISSUE_ID>

## UI changes

## Checklist

<details>

1. type (required)
    - `feat` - A new feature
    - `fix` - A bug fix
    - `chat`
    - `overlay`
    - `shared`
    - `sandbox-overlay`
    - `visualizer-connector`

</details>


apps/chat-api/src/files/files.service.ts
```diff
@@ -242,6 +232,65 @@ export class FilesService extends AppService {
...
+  ): Promise<ListFilesResponseDto> {
+    return this.listFiles(
+      'public',
+      query.path,
+      { ...query, permissions: false },
+      at,
+    );
+  }
+
+  async listSharedFiles(
```
hhharmEpm, line 308: fetchByTab in useDialFileManager.ts never passes limit to any of its listing calls (listFiles, listPublicFiles, listSharedFiles), and the nextToken field from the response is silently discarded. If DIAL Core returns a truncated page, the UI will show only the first page of files with no indication that more exist.

valerydluski: Good catch. The frontend intentionally does not handle file-list pagination, because the BFF is expected to return the complete list to the UI.

I updated the NestJS file listing service so listFiles/listPublicFiles now aggregate all DIAL Core pages when no explicit limit/token is provided by the caller. Explicit paginated requests still preserve the existing limit/token behavior and return nextToken. Added service tests for both cases.

hhharmEpm (approved):

Issue #7512, opened 2026-06-29
Backend-owned chat generation and conversation persistence


stropalov:
## Summary

Currently, chat generation lifecycle and conversation persistence are both managed by the frontend (`useConversationStream.ts`). This causes a state corruption bug: if the user navigates to another conversation while the assistant is still generating, SSE chunks can land in the wrong conversation context and the final `saveConversation()` call can overwrite the wrong path.

This change moves generation ownership and persistence entirely to the backend.

## Key points

- Backend registers each generation by `sessionId + path + generationId`; returns `409` if a concurrent generation is already active on the same path
- Backend assembles the assistant message from SSE chunks server-side (mirrors frontend `applyChunkToMessages` logic for content, attachments, stages, annotations, `form_schema`, `responseId`)
- Backend saves conversation at stream start (user message + empty assistant placeholder), on `[DONE]` (final assembled message), and on stop/error (`wasStoppedByUser` / `hasStreamError` flags)
- Frontend no longer calls `saveConversation()` — on stream complete it reloads the conversation via `getConversation(path)`
- `GenerationContext` lives above the conversation page: SSE streams survive inter-chat navigation within the same tab
- New `POST /api/v1/conversations/completions/stop` endpoint aborts the active stream and triggers a partial save
- `SendCompletionDto` extended with `generationId`, `mode` (`Append` / `ContinueLastUser` / `Regenerate` / `Edit`), optional `messageIndex`
- Stale registry entries (>30 min) are evicted to prevent memory leaks

## New / changed API endpoints

| Endpoint | Change |
|---|---|
| `POST /api/v1/conversations/completions` | Extended with `generationId`, `mode`, `messageIndex`; backend now owns persistence |
| `POST /api/v1/conversations/completions/stop` | New endpoint; aborts active generation, saves partial result |

## Acceptance criteria

- [x] Navigating between conversations during generation does not corrupt conversation state
- [x] Stopping generation persists a partial assistant message with `wasStoppedByUser: true`
- [x] Stream errors persist a partial message with `hasStreamError: true`
- [x] Concurrent generation attempts on the same path return `409`
- [x] Frontend reloads server state on stream complete instead of saving locally

### Confidential information

- [x] I confirm that do not share any confidential information

AnastasiiaHarkot:
Verified in dev

Issue #7436, opened 2026-06-25
Streaming stop causes message role corruption and duplicate state


AnastasiiaHarkot:
### DIAL chat version

1.0.0

### How to reproduce

**Case 1:**
1. Send any message and receive a streaming response
2. Click 'Stop' before generation completes
3. Observe the partial assistant response

**Case 2:**
1. Reproduce Case 1 (stop mid-stream)
2. Click 'Regenerate'
3. Stop again before completion

### Actual result

**Case 1:** The partial response moves into a user-styled bubble with raw markdown syntax visible (##, **, - visible as plain text)
<img width="978" height="781" alt="Image" src="https://github.com/user-attachments/assets/56cd5197-bca0-4ce7-a918-d7ba9ed8c847" />

**Case 2:** Both user messages and assistant messages appear identical and multiplied

<img width="861" height="604" alt="Image" src="https://github.com/user-attachments/assets/d50468a2-e94b-4607-98e9-c8c28ee3ea90" />

### Expected result

**Case 1:** Partial response stays in the assistant bubble, markdown rendered

**Case 2:** One user message, one assistant message with new partial response

### Confidential information

- [x] I confirm that do not share any confidential information

AnastasiiaHarkot:
**What was tested:**
- **Case 1 — Stop mid-stream:** Sent a long essay prompt to Gemini 2.5 Pro (Google Search), waited for the response to start streaming, then clicked the Stop button while the assistant was generating text.
- **Case 2 — Stop → Regenerate → Stop:** After the first stop, clicked Regenerate, waited for a new stream to begin, then clicked Stop again mid-generation.

**What was observed:**
- **Case 1:** The partial response remained in the assistant-styled bubble (AI icon, white background) with markdown rendered as formatted text — no raw `**asterisks**` visible, no role corruption to a user-styled bubble.
- **Case 2:** No message duplication occurred. Confirmed via DOM query that the user message appeared exactly 1 time in the page. An error banner appeared briefly on the second stop (expected behavior for Google Search model), but no extra messages were created.

✅ Bug is confirmed **FIXED** — neither case from the original report reproduces.

---
> 🤖 This verification was performed by **Claude AI** (automated browser testing, not manual QA). Tested on 2026-06-30.

PR #7515, opened 2026-06-29
feat: backend-owned chat generation and conversation persistence (Issue #7512, #7436)

stropalov:
## Description of changes

- Moves chat generation lifecycle and conversation persistence from the frontend to the backend.
- Fixes a state corruption bug where navigating away during streaming could overwrite the wrong
  conversation, and resolves message role corruption on stop.

## Applicable issues

- https://github.com/epam/ai-dial-chat/issues/7512
- https://github.com/epam/ai-dial-chat/issues/7436

## UI changes

## Checklist

- [X ] the pull request name ends with `(Issue #<ISSUE_ID>)` (comma-separated list of issues)
- [X ] I confirm that I don't share any confidential information like API keys or any other secrets and private URLs


stropalov:
/deploy-review

>GitHub actions run: [28376176074](https://github.com/epam/ai-dial-ci/actions/runs/28376176074)

| [deploy-review](https://chat-ai-dial-chat-ng-pr-7515.gke.test.dial.parts) | [Success ✅](https://github.com/epam/ai-dial-ci/actions/runs/28376176074/job/84066295534) |
| matrix.application | [Skipped ➖](https://github.com/epam/ai-dial-ci/actions/runs/28376176074/job/84068547278) |

apps/chat-api/src/conversations/conversation-generation.service.ts
```diff
@@ -0,0 +1,96 @@
+import { ConflictException, Injectable, Logger } from '@nestjs/common';
+
+const STALE_ENTRY_TTL_MS = 30 * 60 * 1000; // 30 minutes
+
+interface GenerationEntry {
+  generationId: string;
+  abortController: AbortController;
+  status: 'active' | 'stopped' | 'done' | 'error';
```
valerydluski: Maybe we should use an enum here instead of inline string literals:
```
enum Status {
  Active = 'active',
  Stopped = 'stopped',
  Done = 'done',
  Error = 'error',
}
```

This would make the statuses easier to reuse and maintain.

apps/chat-api/src/conversations/utils/apply-chunk.server.ts
```diff
@@ -0,0 +1,199 @@
...
+  const result = [...existing];
+  for (const stage of incoming) {
+    const idx = result.findIndex((s) => s.index === stage.index);
+    if (idx >= 0) {
+      result[idx] = {
+        ...result[idx],
+        ...stage,
+        name: (result[idx].name ?? '') + (stage.name ?? ''),
+        content:
+          (result[idx].content ?? '') + (stage.content ?? '') || undefined,
```
valerydluski: **[Required] `|| undefined` silently drops stage content when both deltas are empty strings**

If `result[idx].content` and `stage.content` are both `''`, the concatenation is `''`, which is falsy, so `|| undefined` returns `undefined` — erasing any content that was already set. This fires for header-only or status-only stage frames where `content` is explicitly `''`.

```ts
// Before
content: (result[idx].content ?? '') + (stage.content ?? '') || undefined,

// After
const merged = (result[idx].content ?? '') + (stage.content ?? '');
content: merged !== '' ? merged : undefined,
```

apps/chat-api/src/conversations/conversation-generation.service.ts
valerydluski: **[Required] Use a string enum instead of a string-literal union for `status`**

Per project conventions (AGENTS.md: _"Use string enums for named finite sets of statuses"_), `'active' | 'stopped' | 'done' | 'error'` should be a `GenerationStatus` string enum:

```ts
export enum GenerationStatus {
  Active = 'active',
  Stopped = 'stopped',
  Done = 'done',
  Error = 'error',
}
```

Also note: the frontend `GenerationContext` only has three values (`'active' | 'done' | 'error'`), missing `'stopped'` — the sets should be aligned or the frontend should mirror the backend enum.

apps/chat-api/src/conversations/dto/send-completion.dto.ts
```diff
@@ -1,17 +1,36 @@
...
+}
+
 export class SendCompletionDto {
+  @ApiProperty({
+    description: 'Client-generated UUID identifying this generation attempt.',
+    example: 'cfeaf733-4ecd-4898-ad3b-d6835c0b5fc8',
+  })
+  @IsString()
+  @IsNotEmpty()
+  generationId!: string;
```
valerydluski, line 32: **[Required] `generationId` needs `@IsUUID('4')` validation**

`@IsString() @IsNotEmpty()` accepts any arbitrary string including very long ones. The client is expected to send a UUID v4; enforce it:

```ts
@IsUUID('4')
generationId!: string;
```

This also implicitly caps the length. The `model` field on this DTO similarly has no `@MaxLength()` — it is passed to DIAL Core and logged, so it should be bounded (e.g. `@MaxLength(256)`).

apps/chat-api/src/conversations/dto/stop-completion.dto.ts
```diff
@@ -0,0 +1,23 @@
...
+import { IsNotEmpty, IsString, Matches, MinLength } from 'class-validator';
+
+export class StopCompletionDto {
+  @ApiProperty({
+    description: 'Generation ID that was returned by the active stream.',
+    example: 'cfeaf733-4ecd-4898-ad3b-d6835c0b5fc8',
+  })
+  @IsString()
+  @IsNotEmpty()
+  generationId!: string;
```
valerydluski, line 11: **[Required] `generationId` needs `@IsUUID('4')` validation**

Same issue as `send-completion.dto.ts` — `@IsString() @IsNotEmpty()` accepts any string. Add `@IsUUID('4')` to enforce the expected format and cap length.

apps/chat-api/src/conversations/utils/conversation-history-builder.ts
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  if (mode === CompletionMode.Regenerate) {
+    if (messageIndex == null) {
+      throw new BadRequestException(
+        'messageIndex is required for regenerate mode',
+      );
+    }
+    // Truncate up to (but not including) messageIndex, then append assistant placeholder
+    const truncated = messages.slice(0, messageIndex);
```
valerydluski, line 111: **[Required] No upper-bound check on `messageIndex`**

For `Regenerate` mode, `messageIndex` is validated as non-null but not checked against `messages.length`. A client can send `messageIndex: 9999` on a 2-message conversation; `slice(0, 9999)` silently returns all messages and the regeneration appends to the end rather than replacing the intended position. Same issue exists for `Edit` mode (~line 94).

Add a bounds check after the null check:

```ts
if (messageIndex >= messages.length) {
  throw new BadRequestException(
    `messageIndex ${messageIndex} is out of range for a ${messages.length}-message conversation`,
  );
}
```

A test for out-of-range `messageIndex` in `conversation-history-builder.spec.ts` should accompany this fix (it was listed as a required edge case in the spec tasks).

apps/chat-api/src/conversations/conversation.controller.ts
```diff
@@ -196,76 +199,62 @@ export class ConversationController {
...
-    } finally {
-      if (keepaliveTimer) clearInterval(keepaliveTimer);
-      res.off('close', handleClose);
-      isReaderReleased = true;
-      reader.releaseLock();
-      if (!res.writableEnded) {
-        res.end();
-      }
+  @Post('completions/stop')
+  @HttpCode(204)
```
valerydluski: **[Nit] `@HttpCode(204)` is dead code here**

When `@Res()` is injected into a handler, NestJS bypasses its response lifecycle — including the `@HttpCode` decorator. The 204 status is set by `res.status(204).end()` below. The decorator has no effect and misleads readers. Consider removing it.

apps/chat/src/context/GenerationContext.tsx
```diff
@@ -0,0 +1,97 @@
...
+  status: 'active' | 'done' | 'error';
+}
+
+interface Props {
+  children: ReactNode;
+}
+
+interface GenerationContextValue {
+  startGeneration: (path: string, generationId: string) => AbortController;
+  stopGeneration: (path: string, generationId: string) => void;
```
valerydluski: **[Nit] `stopGeneration` is exported in context but never consumed**

`GenerationContextValue.stopGeneration` is implemented and spread into the context value, but no consumer destructures it — `handleStop` in `useConversationStream.ts` calls `stopCompletion` (the HTTP API) directly without going through `stopGeneration`. This leaves dead public API in the context shape.

Either remove it to keep the context minimal, or add a brief comment explaining it is reserved for future use.

apps/chat/src/server-api/chat-stream.api.ts
```diff
@@ -1,20 +1,43 @@
...
   onComplete: () => void;
   onError: (error: Error) => void;
   signal?: AbortSignal;
 }
 
+export const stopCompletion = async (dto: {
+  generationId: string;
+  path: string;
+}): Promise<void> => {
+  await fetch(`${ApiEndpoints.CONVERSATIONS}/completions/stop`, {
```
valerydluski: **[Nit] `stopCompletion` swallows fetch errors silently**

If the stop request returns 404, 500, or fails with a network error, the error is lost. The user gets no feedback that the stop did not work and the stream continues.

Surface the error to the caller:

```ts
const res = await fetch(`${ApiEndpoints.CONVERSATIONS}/completions/stop`, { ... });
if (!res.ok) {
  throw new Error(`stopCompletion failed: ${res.status}`);
}
```

The call site in `useConversationStream.ts` can then log or surface it via `onError`.

valerydluski:
**[Note] `streamCompletion` has 11 positional parameters — consider an options object** (apps/chat-api/src/conversations/conversation.service.ts)

The 11-argument signature makes call sites fragile (argument order, optional params). The integration test already uses placeholder names like `_path, _at, _bucket, _genId, _mode, _msg, _msgIdx, _model, _cc, _sid, res` to track position.

Grouping into an options object would make call sites self-documenting:

```ts
interface StreamCompletionOptions {
  conversationPath: string;
  token: string;
  bucket: string;
  generationId: string;
  mode: CompletionMode;
  message?: string;
  messageIndex?: number;
  model: string;
  customContent?: MessageCustomContentDto;
  sessionId: string;
}

async streamCompletion(options: StreamCompletionOptions, res: Response): Promise<void>
```

Not blocking for this PR, but worth addressing before the signature grows further.

valerydluski (dismissed):

PR #7569, opened 2026-07-01
feat: implement inline comment publishing for findings in PRs

valerydluski:
## Applicable issues

- fixes #<ISSUE_ID>

## UI changes

## Checklist

<details>

1. type (required)
    - `feat` - A new feature
    - `fix` - A bug fix
    - `chat`
    - `overlay`
    - `shared`
    - `sandbox-overlay`
    - `visualizer-connector`

</details>

### DIFF
--- a/.github/claude/scripts/publish-inline-comments.py
+++ b/.github/claude/scripts/publish-inline-comments.py
@@ -9,26 +9,15 @@
 
 import hashlib
 import json
-import math
 import os
 import re
 import subprocess
 import sys
-import urllib.parse
-from collections import Counter
 from pathlib import Path
 
+sys.path.insert(0, str(Path(__file__).parent))
+from sdlc_scrub import neutralize_public_body  # noqa: E402
 
-SECRET_PATTERNS = {
-    "anthropic-key": r"sk-ant-[A-Za-z0-9_\-]{20,}",
-    "openai-style-key": r"\bsk-[A-Za-z0-9]{32,}\b",
-    "aws-access-key": r"\b(?:AKIA|ASIA)[0-9A-Z]{16}\b",
-    "github-token": r"\bgh[pousr]_[A-Za-z0-9]{36,}\b",
-    "github-fine-pat": r"\bgithub_pat_[A-Za-z0-9_]{60,}\b",
-    "slack-token": r"\bxox[baprs]-[A-Za-z0-9-]{10,}\b",
-    "jwt": r"\beyJ[A-Za-z0-9_\-]{10,}\.eyJ[A-Za-z0-9_\-]{10,}\.[A-Za-z0-9_\-]{10,}\b",
-    "private-key-block": r"-----BEGIN (?:[A-Z ]+ )?PRIVATE KEY-----",
-}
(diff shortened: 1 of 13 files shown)


DmytroZaichenkoDev (commented):
Reviewed the inline-comment publishing. Mechanics look solid (diff line-mapping, fail-closed secret gate, idempotency, correct opt-out for security-review). One thing worth addressing before merge (duplicated M4 scrubber) plus two smaller notes inline.

.github/claude/scripts/publish-inline-comments.py
```diff
@@ -0,0 +1,280 @@
...
+import os
+import re
+import subprocess
+import sys
+import urllib.parse
+from collections import Counter
+from pathlib import Path
+
+
+SECRET_PATTERNS = {
```
DmytroZaichenkoDev: This duplicates the secret-scan + URL/HTML neutralizer from `scrub-output.py`. Since inline-posted findings are dropped from the sticky body, they're gated *only* by this copy — so if `scrub-output.py` ever gains a pattern, inline comments silently bypass it. Suggest extracting a shared module both call.

.github/claude/scripts/render-stage-comment.py
```diff
@@ -39,24 +40,50 @@ def deep_merge(base, override):
     return out
 
 
+def finding_message(finding):
```
DmytroZaichenkoDev, line 43: This changes the sticky table for all findings agents, not just review: it now concatenates message/title/description/details/reason/evidence and appends `suggested_fix` into the Message cell. Intended? It's a repo-wide rendering change riding along with a review-only feature.

.github/claude/scripts/publish-inline-comments.py
```diff
@@ -0,0 +1,280 @@
...
+
+    files = run_gh_paginated(f"repos/{repo}/pulls/{pr_number}/files")
+    right_lines_by_path = {
+        item.get("filename"): parse_right_lines(item.get("patch") or "")
+        for item in files
+        if item.get("filename")
+    }
+
+    existing_comments = run_gh_paginated(f"repos/{repo}/pulls/{pr_number}/comments")
+    existing_markers = {
```
DmytroZaichenkoDev, line 127: `existing_markers` isn't updated as comments post in the loop, so two findings in one run that hash to the same marker would both post. Minor — `existing_markers.add(marker)` after a successful post closes it.

.github/claude/scripts/render-stage-comment.py
valerydluski: Intentional. `finding_message()` was introduced for inline comments and reused in the sticky table on purpose — previously only the `message` field appeared in the Message column, silently discarding `title`, `description`, `reason`, and other fields agents populate instead. Unifying on the same helper makes both views consistent.

.github/claude/scripts/publish-inline-comments.py
valerydluski: fixed

.github/claude/scripts/publish-inline-comments.py
valerydluski: fixed

PolinaGurinovich97 (approved):

PR #7590, opened 2026-07-03
ci: make security review incremental and memory-aware

dearsimanovich-epam:
## Problem

The Claude Security Review workflow ran an open-ended "find problems" pass over the **whole** PR diff on every push (`pull_request: synchronize`). Being stateless and non-deterministic, each run sampled a different subset of possible findings — fix the flagged items and the next run surfaced a fresh, unrelated batch. The review never converged.

## Fix (single file: `.github/workflows/claude-security-review.yml`)

1. **Incremental** — review only `git diff PREV_HEAD_SHA..HEAD_SHA` (the newly pushed commits, via `github.event.before`/`after`). Falls back to the full `base...head` diff on first review, force-push, or rebase, detected with `git merge-base --is-ancestor`. If it never re-looks at unchanged code, it structurally cannot invent new comments about it.
2. **Memory-aware** — loads its own prior inline comments and thread resolution status via `gh api graphql`, then omits resolved findings, re-checks open ones, and never repeats itself.
3. **Severity gate** — reports only Medium+ security/correctness issues; forbids style/naming/nitpick comments. Marks the review advisory (non-blocking) in the sticky comment.
4. Allowlist gains `git merge-base` and `gh api graphql` (scoped, read-only).

Stays **security-scoped** — same `security-review` skill and `security-review` label gate.

## Verify (on a throwaway PR labeled `security-review`)

1. First push → full-diff review + sticky comment.
2. Push a fix → re-run touches only the new commit, resolves/omits the fixed finding, posts no unrelated comments.
3. Push a whitespace-only commit → zero findings outside that file = loop broken.
4. Rebase + push → falls back to full diff without error; memory-aware dedupe avoids duplicates.


stropalov (approved):

valerydluski (approved):

dearsimanovich-epam:
Reverting: the endless-review loop is in the framework code-review agent, not this standalone security workflow (which was skipped on the affected PRs). Fix will target the code-review agent path without touching the security agent.

Issue #7643, opened 2026-07-08
New toolset not shown in Catalog until page reload


valerydluski:
### EPAM AI DIAL Admin version

0.0.1

### How to reproduce

1. Open the Toolset Editor.
2. Create and save a new toolset.
3. Return to the Catalog view.
4. The newly created toolset does not appear in the list.
5. Reload the page — the toolset now appears.

### Actual result

After creating a new toolset, it does not appear in the Catalog list.

### Expected result

The newly created toolset appears in the Catalog list immediately after creation, without requiring a page reload.

### Details

- `ToolsetEditor.tsx` `handleSave` (apps/chat/src/pages/ToolsetEditor/ToolsetEditor.tsx:266-272) calls `createToolset()` and then only runs `runPostSaveAuth` + `navigate(returnUrl)` — it never triggers a refetch or updates the catalog list state.
- Catalog data (`toolsets`) lives in `DeploymentsContext` (apps/chat/src/context/DeploymentsContext.tsx) and is loaded once via `loadDeployments` inside a `useEffect` (lines 173-179) keyed only on `userConfigSelectedId` / `appConfig.defaultDeploymentId` — it never reruns after a toolset is created.
- `DeploymentsContextType` (lines 24-47) exposes no `refetch`/`reload` function, so there's no mechanism to invalidate or append to the list after a mutation.
- A full page reload remounts `DeploymentsProvider`, re-running the effect and picking up the new toolset — which is why reload "fixes" it.

### Confidential information

- [x] I confirm that do not share any confidential information

AnastasiiaHarkot:
Verified on the current build. Created a new toolset "test-qa-7643" via Catalog → Create → Toolset, filled in a name and endpoint, then clicked Save & Exit. The app navigated back to the Catalog and the new toolset appeared immediately in the Toolsets tab (count updated from 191 to 192) without any page reload. The bug is no longer reproducible.

**Verdict: FIXED** — closing this issue.

---
*This verification was performed by **Claude AI** (automated browser testing, not manual QA).*

PR #7648, opened 2026-07-08
fix: refresh catalog toolset list after create/update a toolset (#7643)

stropalov:
## Description of changes

- Newly created/updated toolsets didn't appear in the Catalog until a full page reload, because `DeploymentsContext` loaded toolsets once on mount with no way to re-fetch them.
- Added `refetchToolsets()` to `DeploymentsContext` and call it from `ToolsetEditor.handleSave` right after a successful create/update, before navigating back to the catalog.

## Applicable issues

- https://github.com/epam/ai-dial-chat/issues/7643

## Checklist

- [X] the pull request name ends with `(Issue #<ISSUE_ID>)` (comma-separated list of issues)
- [X] I confirm that I don't share any confidential information like API keys or any other secrets and private URLs

### DIFF
--- a/apps/chat/src/context/DeploymentsContext.tsx
+++ b/apps/chat/src/context/DeploymentsContext.tsx
@@ -44,6 +44,8 @@ export interface DeploymentsContextType {
   schemas: ApplicationSchemaSummaryDto[];
   /** Toolsets fetched from the dedicated toolsets API for catalog surfaces. */
   toolsets: DialToolsetDto[];
+  /** Re-fetches toolsets from the API and updates the catalog list. Call after creating/updating a toolset. */
+  refetchToolsets: () => Promise<void>;
 }
 
 export const DeploymentsContext = createContext<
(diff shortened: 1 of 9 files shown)


apps/chat/src/context/DeploymentsContext.tsx
```diff
@@ -178,6 +180,15 @@ export const DeploymentsProvider = ({ children }: { children: ReactNode }) => {
     };
   }, [loadDeployments]);
 
+  const refetchToolsets = useCallback(async () => {
+    try {
+      const { data } = await listToolsets();
+      setToolsets(sortToolsets(data ?? []));
+    } catch (err) {
+      console.warn('[DeploymentsContext] Failed to refetch toolsets', err);
```
valerydluski: Let's remove the console log and add a toast notification instead.

stropalov: Done

Issue #7783, opened 2026-07-14
[Quick App] Conversation Starters & Intro Text Not Displayed in Preview/App


Serg0147:
### DIAL chat version

1.0.0

### How to reproduce

1. Open DIAL Chat on Desktop
2. Create a new Quick App 2.0
3. Go to Settings → Conversation Starters
4. Add starters e.g.: hi / hi
5. Add Intro text: "this is a new text"
6. Click Preview OR Save & Exit and open the app
7. Observe the app/preview screen

### Actual result

❌ No conversation starters shown in preview/app
❌ No intro text displayed
❌ Only empty chat with "Send a message to preview this application..." shown

<img width="2966" height="1153" alt="Image" src="https://github.com/user-attachments/assets/a4efeb79-668a-477c-b784-66c90bc2e8d4" />

<img width="2830" height="1216" alt="Image" src="https://github.com/user-attachments/assets/95084907-1ab9-495d-beb6-06bdea879427" />

### Expected result

Conversation starter buttons visible near chat input
Intro text displayed above chat input

<img width="628" height="207" alt="Image" src="https://github.com/user-attachments/assets/3eadff9c-74f6-42d2-9399-14e87b143513" />

### Confidential information

- [x] I confirm that do not share any confidential information

natshabat:
fixed 

<img width="1851" height="452" alt="Image" src="https://github.com/user-attachments/assets/e0186932-a03e-4b79-8fce-9349c9794b27" />


<img width="853" height="297" alt="Image" src="https://github.com/user-attachments/assets/47a35214-a77b-480a-81ff-882c33892a64" />

PR #7874, opened 2026-07-17
fix: show conversation starters and intro text (#7783)

stropalov:
### Summary

Fixed Quick Apps conversation starters and intro text not appearing in both the main new-chat screen and Apps Editor Preview.

### Root Cause

Quick Apps settings store conversation starter data under `application_properties.conversation_starters`, but the deployments list API was not exposing that data to the frontend. The frontend only handled schema-based starters, so Quick Apps `intro_text`, starter buttons, `auto_submit`, and `chat_message_input_disabled` were ignored.

Preview also reused the deployments list after saving settings, but that list could still be served from the short deployments cache, so newly saved settings were not reflected immediately.

## Applicable issues

- https://github.com/epam/ai-dial-chat/issues/7783

## Checklist

- [X] the pull request name ends with `(Issue #<ISSUE_ID>)` (comma-separated list of issues)
- [X] I confirm that I don't share any confidential information like API keys or any other secrets and private URLs

### DIFF
--- a/apps/chat/src/components/NewConversationComposer/NewConversationComposer.tsx
+++ b/apps/chat/src/components/NewConversationComposer/NewConversationComposer.tsx
@@ -70,6 +70,8 @@ interface Props {
   modelSelectorError?: unknown;
   isInputDisabled?: boolean;
   placeholder: string;
+  /** Optional text shown above the starter buttons and the composer input. */
+  introText?: string;
   /** Initial textarea content (e.g. populated by a starter selection). */
   message?: string;
   inputStyles?: ConversationInputStyles;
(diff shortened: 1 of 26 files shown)


apps/chat/src/components/NewConversationComposer/NewConversationComposer.tsx
```diff
@@ -256,6 +259,12 @@ const NewConversationComposer: FC<Props> = ({
         role="region"
         aria-label={t(ChatI18nKeys.WelcomeScreen)}
       >
+        {introText && (
+          <p className="mb-4 max-w-3xl text-center text-sm text-secondary">
```
hhharmEpm: can we use dial system typography here? dial-small-text?

stropalov: Sure, done.

hhharmEpm (approved):

Issue #3073, opened 2025-01-31
[Apps]: Incorrect size of drop-down icon for "Topics" and  and color of x-icon for "Attachment types" fields


YauheniyaH:
### EPAM AI DIAL chat version

0.24.0

### How to reproduce

drop-down icon for "Topics" has size 20x20 (should be 18px)
 x-icon for "Attachment types" has color #F3F4F6 ( should be #7F8792)

### Actual result

<img width="854" alt="Image" src="https://github.com/user-attachments/assets/30e2c6ca-f87c-4d4e-84ec-4e90bff4da60" />

![Image](https://github.com/user-attachments/assets/57db72fe-ae47-4145-a7a5-129d9cf40a6b)

<img width="752" alt="Image" src="https://github.com/user-attachments/assets/3ff70582-e7c5-4496-9838-f5d589536b4f" />

### Expected result

https://www.figma.com/design/lwL5jneveZczAj5RLcLFH2/DIAL-Chat?node-id=12044-75703&t=b6VHCrAJKqau9G3G-4

### Confidential information

- [x] I confirm that do not share any confidential information

PolinaGurinovich97:
Blocked - waiting new Select

VolhaBazhkova:
Verified in  DIAL Chat Version 0.49.0-dev.57. @PolinaGurinovich97 could you please implement the following:
A) 'x' button in the element selected in drop down list should be from ui-kit (the size is 16px -> ok, but on hover over there is no square effect -> bug). In DIAL there are two places:
1. While creation/editing any application: **Topics**
2. While creation/editing toolset, e.g. Notion: **Allowed tools**

<img width="1151" height="155" alt="Image" src="https://github.com/user-attachments/assets/2451bd55-59b1-427f-a3d2-b8e010caae5a" />

<img width="799" height="129" alt="Image" src="https://github.com/user-attachments/assets/f8a5b3a1-4f42-46c6-a9ea-1bb02b081798" />

B) in the same drop down lists now there are elements x and arrow with different sizes. 
Expected:
1. 'x' icon should be from ui-kit => should have size 16px and have square effect on hover over
2. arrow icon should not have effect on hover over (now it works so -> ok), but the size should be 16px (now it's 18px or 20px -> bug)

VolhaBazhkova:
Verified in DIAL Chat Version 0.49.0-dev.78
The buttons and the sizes are fine now. 

[...]

PR #8286, opened 2026-08-11
fix(chat): attachment types MIME error and attachment name centering (Issue #5256, Issue #3073)

### LINKED ISSUE #5256: [App editor] Incorrect behavior when there is invalid value set in 'Attachment types' field
### DIAL chat version

0.41.0

### What happened?

1. Start custom app creation
2. On the second step ('App settings') type incorrect value for 'Attachment types' -> red 'Please match the MIME format' error message appears (ok)
3. Type any text into 'Max. attachments number' -> red 'Invalid input: expected number, received NaN' error message appears (ok)
4. Move the mouse on Preview area -> red error in 'Attachment types' disappears -> **bug1** 
It seems that sometimes it's enough to move the mouse out of  'Max. attachments number' to have the error in  'Attachment types' disappears/appears

**Note1:** also incorrect value and the error in 'Attachment types' disappears when user returns to 'General info' -> **bug2**
**Note2:** when there is incorrect 'Attachment types'  and user clicks on 'Save and exit' -> there is no 'Only valid data will be saved' confirmation pop-up -> **bug3**
**Note3:** the issue is reproduced for in Quick app 2.0 also (we need to re-check all the app types with attachments)

Image

https://github.com/user-attachments/assets/e79f12e9-c5a7-4a14-b82c-379b1a380d39

**Expected:** the error message should stay if the format is incorrect

### Confidential information

- [x] I confirm that do not share any confidential information

PolinaGurinovich97:
## Description of changes

Two fixes in the attachment fields.

### The attachment types MIME error is not shown while typing (#5256)

Fixes the regression reported in [this comment](https://github.com/epam/ai-dial-chat/issues/5256#issuecomment-5255230901): typing an invalid attachment type and leaving the field no longer showed `Please match the MIME format`.

The previous fix for #5256 replaced the manual `setError` with a schema rule, so the message would survive form re-validation and switching editor steps. The rule validates the typed but not yet added value (`pendingInputAttachmentType`) and reports on `inputAttachmentTypes`, because that is the field the message belongs to. Typing calls `setValue('pendingInputAttachmentType', …, { shouldValidate: true })`, and react-hook-form refreshes the errors of the changed field only — an error reported on another field never reached `formState.errors` until the whole form was validated. Hence no message while typing or on blur.

The field now derives its message from the typed value directly, via `getPendingAttachmentTypeError`, instead of reading it from the form errors. Nothing else changes:

- the schema keeps the same rule, so the form stays invalid, `Save and exit` still shows the `Only valid data will be saved` dialog, and the invalid state still maps to the `App settings` step;
- the typed value still lives in the form state, so it and its message survive switching to `General info` and back and `Continue editing` in the dialog — the behaviour the original issue asked for.

Both the schema rule and the field message go through the same helper, so they cannot disagree.

### The attachment name is not centered in the input chips (#3073)

Requested in [this comment](https://github.com/epam/ai-dial-chat/issues/3073#issuecomment-5254498648): in the send message input and in a message opened for editing, the remove button of an attachment chip is centered while the file name sits above the center.

The row that holds the name and the remove button had no cross axis alignment, so both were stretched to the row height. The button fills its own box and looks centered, while the name column starts at the top of the stretched row. The row items are now centered, in the file, folder and link chips, which all share that markup.

## Applicable issues

- fixes #5256
- part of #3073

## UI changes

The `Please match the MIME format` message appears again as soon as the typed value is not a MIME type, in the custom app, code app and quick app 2.0 editors. Attachment names are vertically centered in the input chips.

## Checklist

- [x] the pull request name ends with `(Issue #<ISSUE_ID>)` (comma-separated list of issues)

### DIFF
--- a/apps/chat/src/components/AppsEditor/EditorForm/CodeAppForm.tsx
+++ b/apps/chat/src/components/AppsEditor/EditorForm/CodeAppForm.tsx
@@ -32,6 +32,7 @@ import {
 } from '@/src/constants/applications';
 import { CODE_APPS_ENDPOINTS } from '@/src/constants/code-apps';
 import { CommonI18nKeys, MarketplaceI18nKeys } from '@/src/constants/i18n';
+import { getPendingAttachmentTypeError } from '@/src/constants/validation-helpers';
 
 import {
   CodeAppForm as CodeAppFormType,
@@ -149,7 +150,10 @@ export const CodeAppForm = () => {
             hasDeleteAll
             hideSuggestions
             itemHeightClassName="h-[31px]"
-            error={errors.inputAttachmentTypes?.message}
+            error={
+              errors.inputAttachmentTypes?.message ??
+              getPendingAttachmentTypeError(pendingAttachmentType)
+            }
             disabled={isAppPublic}
             tooltip={isAppPublic ? PUBLIC_APP_TOOLTIP : ''}
             {...getPendingAttachmentTypeProps(pendingAttachmentType, setValue)}
--- a/apps/chat/src/components/AppsEditor/EditorForm/CustomAppForm.tsx
+++ b/apps/chat/src/components/AppsEditor/EditorForm/CustomAppForm.tsx
@@ -21,6 +21,7 @@ import { ApplicationSelectors } from '@/src/store/selectors';
 
 import { PUBLIC_APP_TOOLTIP } from '@/src/constants/applications';
 import { CommonI18nKeys, MarketplaceI18nKeys } from '@/src/constants/i18n';
+import { getPendingAttachmentTypeError } from '@/src/constants/validation-helpers';
 
 import {
   CustomAppForm as CustomAppFormType,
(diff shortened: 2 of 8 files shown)


PolinaGurinovich97:
/deploy-review

>GitHub actions run: [31532450769](https://github.com/epam/ai-dial-ci/actions/runs/31532450769)

| [deploy-review](https://chat-ai-dial-chat-pr-8286.gke.test.dial.parts) | [Success ✅](https://github.com/epam/ai-dial-ci/actions/runs/31532450769/job/93915513171) |
| chat | [Success ✅](https://github.com/epam/ai-dial-ci/actions/runs/31532450769/job/93919999517) |
| overlay | [Success ✅](https://github.com/epam/ai-dial-ci/actions/runs/31532450769/job/93919999547) |

denys-kolomiitsev (approved):

PR #8297, opened 2026-08-12
feat(catalog): add Prompt entity type with editor, favorites, and sharing

PolinaGurinovich97:
## Description of changes

Makes prompts first-class citizens of the unified catalog: searchable and sortable alongside models, agents, and toolsets, readable in the details panel, insertable into the composer in one click, and fully authorable through a new editor route. Everything is behind `OverlayFeature.Prompts`.

Spec: `openspec/changes/add-prompt-catalog-entity-type/` (proposal, design with 14 decisions, 11 capability specs, tasks).

### `libs/catalog`

- `CatalogEntityType.Prompt` with its entity colour, tab label, and tab position after Toolsets.
- An optional `promptContent` details field backing a **Details** tab that renders the prompt's description above its body. The body is markdown, and each `{{placeholder}}` is wrapped in `span.cat-prompt-variable`. The lib emits the class and **no colour for it** — styling that class is the host's, so nothing here paints it yet.
- A prompt shows exactly two tabs, Details then Overview, and never About: the description already sits in Details and the storage metadata in Overview, so About would only repeat them. Every other entity type is untouched.
- `isUnshareVisible`, so a host can declare unsharing unsupported per item.

### `libs/prompt-editor` (new)

The prompt form and its folder picker, deliberately passive: no validation against a storage contract, no API call, no route, no i18n. Messages arrive through `errors`, folder mutations through `folderActions`, and `onValidateFolderName` keeps the name pattern outside the lib. That leaves the same form usable by a host whose prompt storage differs from DIAL's.

### `apps/chat`

- `prompts.api.ts` over all eleven generated `PromptsApi` methods; `PromptsContext` owning personal, shared, and organisation prompts.
- Prompt items merged into `CatalogView`, with type-dispatched details fetch, delete, edit, and use-in-chat.
- Use-in-chat hands the body to the composer as one-shot router state rather than a query param, which a 50k-character body would overflow.
- The `PromptEditor` route as a thin container over the new lib.

### `apps/chat-api`

- **Prompt favourites** — a `prompts.installed` user-config section, config version 3 → 4, and `PATCH /api/v1/user-config/prompts`. It needs its own DTO: a prompt path legitimately contains spaces, which `UpdateInstalledDto`'s no-whitespace rule rejects.
- **Prompt share links** — `CreateShareLinkDto.resourceKind` lets the service qualify a bucket-relative prompt path with the caller's own bucket, so the bucket never has to reach the client. An explicit discriminator rather than prefix sniffing, which cannot tell a bare `AI/summarize` from an already-qualified path whose bucket happens to be `AI`.
- **DIAL folder markers** — `.dial_folder` no longer surfaces as a broken prompt in any listing, matching the conversations, deployments, and toolsets services.

### Worth a reviewer's attention

- **A pre-existing bug is fixed here.** Regenerating the client showed `PublishRuleDto` declared `_function` while the swagger and the wire say `function`. Bodies are `JSON.stringify`'d as-is, so publish-with-rules was sending a key the backend's `forbidNonWhitelisted` validator rejects. Fixed in `publish-rules.api.ts`.
- **`prompts` is in `DEFAULT_ENABLED_UI_FEATURES`**, so this ships enabled rather than dark as the design's rollout plan assumed. Deliberate and flagged in the design doc; removing it from the default set restores the dark rollout.
- **Scope grew past the original plan.** The change first forbade backend edits and documented favourites, sharing, and marker filtering as limitations; all three were then pulled in. The design doc records each revision instead of rewriting history.
- **Three test failures on this branch are pre-existing**, reproduced with the whole change stashed: the `unshare` block in `CatalogView.spec.tsx` expects `variant: 'success'` where the code emits `'info'`.
- `openapi.json` needs a Prettier pass after `npm run openapi` — the generator writes it unformatted, which otherwise shows up as ~1500 lines of whitespace diff.

### Still to do

Tasks 8.3–8.6: the RTL (Arabic), a11y keyboard, and mobile passes, plus the five-axis review. All need a browser.

### Verification

| Suite | Result |
| --- | --- |
| `chat-api` | 2130 passed |
| `apps/chat` | 2500 passed (3 pre-existing `unshare` failures) |
| `libs/catalog` | 276 passed |
| `libs/prompt-editor` | 27 passed |
| `libs/chat-shared` | 195 passed |
| `libs/chat-overlay` | 72 passed |

Lint and build green for every touched project; `openapi:check` clean; `tsc --noEmit` clean for the app and spec configs.

## Applicable issues

No linked issue — this was requested directly, so the title carries no `(Issue #N)` suffix.

## UI changes

The prompt details panel now shows **Details** (description + markdown body with highlighted `{{placeholders}}`) and **Overview** (folder, source, last updated). Screenshot to be attached.

## Checklist

- [ ] the pull request name ends with `(Issue #<ISSUE_ID>)` (comma-separated list of issues) — n/a, no linked issue


apps/chat/src/components/SharePopoverContainer/SharePopoverContainer.tsx
```diff
@@ -37,7 +38,18 @@ interface Props {
  */
 const SharePopoverContainer: FC<Props> = ({ item, onClose }) => {
   const { t } = useTranslation();
-  const { data, isLoading, error, setAccess } = useShareLink(item.id);
+  /*
+   * A prompt's `item.id` is a bucket-relative path, so the backend has to be
+   * told which namespace to qualify it against.
+   */
+  const resourceKind =
```
PolinaGurinovich97, line 45: check

apps/chat/src/pages/PromptEditor/PromptEditor.tsx
```diff
@@ -0,0 +1,378 @@
+import { OverlayFeature } from '@epam/ai-dial-chat-overlay';
```
PolinaGurinovich97, line 1: @valerydluski  CC

apps/chat/src/types/prompt-editor.ts
```diff
@@ -0,0 +1,7 @@
+/** Query params the prompt editor route reads. */
+export enum PromptEditorQuery {
```
PolinaGurinovich97, line 2: need?

apps/chat/src/types/prompt.ts
```diff
@@ -0,0 +1,21 @@
+/** Which prompt namespace a catalog prompt item came from. */
```
PolinaGurinovich97, line 1: review file

apps/chat/src/utils/prompt.ts
```diff
@@ -0,0 +1,66 @@
+import { PromptFieldError } from '../types/prompt';
```
PolinaGurinovich97, line 1: review file

libs/catalog/src/components/Details/TabsContent/Content.tsx
```diff
@@ -0,0 +1,61 @@
+import { MarkdownRenderer, mergeClasses } from '@epam/ai-dial-chat-shared';
```
PolinaGurinovich97, line 1: Palina - review

libs/catalog/src/models/catalog-props.ts
```diff
@@ -142,12 +142,12 @@ export interface CatalogProps {
...
-   * step, for an item the current user owns (`isMyApp: true`), removing every
-   * recipient's shared access at once. May return a promise; the confirmation
-   * shows a loading state and prevents duplicate submission while pending.
+   * Narrows where the "Remove from My List" action is offered, on top of the
+   * built-in `sharedWithMe`/`isMyApp` rule. Use it to declare that unsharing
+   * is unsupported for a kind of item. Defaults to `true` (visible) when
+   * absent.
    */
-  onRevokeShare?: (item: CatalogItem) => Promise<void> | void;
+  isUnshareVisible?: (item: CatalogItem) => boolean;
```
PolinaGurinovich97, line 150: check

libs/catalog/src/utils/prompt-variables.ts
```diff
@@ -0,0 +1,80 @@
...
+  children?: HastNode[];
+}
+
+/*
+ * `code` and `pre` are skipped: inside a fenced block a placeholder is being
+ * shown as literal syntax, so highlighting it would misrepresent it as live.
+ */
+const OPAQUE_TAG_NAMES = new Set(['code', 'pre']);
+
+const splitTextNode = (value: string): HastNode[] | null => {
```
PolinaGurinovich97, line 26: review

Issue #8505, opened 2026-08-30
Toolset sign-in never requests offline usage consent, so background use still fails


illya-havsiyevych:
### DIAL chat version

1.0.0-rc.7 (also present on development @ 576e1b04)

### How to reproduce

1. Configure an application that uses a toolset while the user is NOT present
   (a scheduled task, or any work outliving the completion's per-request key).
2. Sign in to that toolset from DIAL Chat - the sign-in interrupt dialog, the toolset
   editor, or the marketplace.
3. Observe the toolset reports SIGNED_IN.
4. Let the application perform its on-behalf-of (OBO) mint for that toolset.

### Actual result

The OBO mint fails with `consent-required` although the user signed in successfully.

This is the same defect as #8504, on the toolset half of the same Core contract.
`offlineUsageConsent` is a field of DIAL Core's shared `ResourceSignInRequest`, which
backs `POST /v1/ops/toolset/signin` (`ResourceCredentialsController`) exactly as it backs
the external-service sign-in. Chat sets it on neither - the string
`offline_usage_consent` / `offlineUsageConsent` appears nowhere in `apps/` or `libs/`.

So a toolset signed in from Chat works while the user watches and stops working the moment
an application reaches for it in the background, with nothing in the UI to explain why.

Filed separately from #8504 because it is a distinct sign-in path
(`useToolsetLogin` / `ToolsetLoginBodyDto`), though the fix can share the same consent
control in the sign-in interrupt dialog.

### Expected result

Let the user grant offline usage consent for a toolset too, and send it with the sign-in.

As in #8504: ask rather than assume - the grant should be visible and declinable rather than
attached silently to a login - and nothing should be defaulted server-side, so the request
carries exactly what the user chose, including an explicit `false`.

Because a single sign-in interrupt can list a toolset and an external service together, one
consent control covering both reads better than two: it is the same question about the same
kind of permission.

I have a PR ready implementing this, stacked on the #8504 PR since they share that control.

One note for whoever picks this up: the frontend toolset path submits through the generated
API client, whose `ToolsetLoginBodyDto` needs the new optional field. `npm run openapi` is
currently blocked by a pre-existing, unrelated bug in `app-config.service.ts` (already noted
in `apps/chat/src/server-api/external-services.ts`), so the PR adds the field to the
generated model by hand, matching what regeneration would emit. Worth regenerating once that
bug is fixed.

### Confidential information

- [x] I confirm that do not share any confidential information

[...]

Issue #8504, opened 2026-08-30
Chat never requests offline usage consent, so a signed-in service still fails on-behalf-of use


illya-havsiyevych:
### DIAL chat version

1.0.0-rc.7 (also present on development @ 576e1b04)

### How to reproduce

1. Configure an application that acts on a user's behalf while the user is NOT present
   (a scheduled run, or any work that outlives the completion's per-request key).
2. Sign in to one of its external services from DIAL Chat - the sign-in interrupt dialog,
   or any other Chat sign-in path.
3. Observe the service now reports `userLevelAuthStatus: SIGNED_IN`.
4. Let the application perform its on-behalf-of (OBO) mint for that same service
   (`POST /v1/ops/external-service/obo-credentials`).

### Actual result

The OBO mint fails with `consent-required` even though the user has just signed in
successfully and the service reports SIGNED_IN.

Cause: DIAL Core records `offline_usage_consent` on a credential and gates every
on-behalf-of mint on it. Chat never sends the flag - the string
`offline_usage_consent` / `offlineUsageConsent` appears nowhere in `apps/` or `libs/`.
Chat's external-service sign-in body is
`{url, credentialsLevel, authenticationType, apiKey, code, redirectUri}`.

The user is left with no way to reason about it: the sign-in succeeded, the UI says
SIGNED_IN, and the application still cannot use the credential. The only remedy today is
to sign in to the same service again from a different product's UI that does send the flag.

Note the same gap exists for toolsets: `offlineUsageConsent` is a field of Core's shared
`ResourceSignInRequest`, which also backs `POST /v1/ops/toolset/signin`. Filed separately.

### Expected result

Let the user grant offline usage consent from Chat, and send it with the sign-in.

Two things matter about how:

- Ask, do not assume. Clicking "Log in" to continue a conversation is not by itself
  agreement that an application may act for the user while they are away, so the grant
  should be visible and declinable rather than attached silently to the login.
- The sign-in interrupt is a natural place to offer it, because that prompt exists
  precisely because an application has said it needs to act on the user's behalf.

Nothing should be defaulted server-side: the request body should carry exactly what the
user chose, including an explicit `false`.

I have a PR ready implementing this (checkbox in the sign-in interrupt dialog, value carried
through the OAuth popup redirect state so the callback can submit it after the round-trip),
with unit tests.

### Confidential information

- [x] I confirm that do not share any confidential information

YuriiDavydiak:
tested on dev
fixed

[...]

PR #8506, opened 2026-08-30
feat(chat): request offline usage consent when signing in to an external service (Issue #8504)

illya-havsiyevych:
**Description:**

DIAL Core records `offline_usage_consent` on a credential and gates every on-behalf-of mint on it. Chat never sent the flag, so an application that works while its user is away could be signed in from Chat and still not run: the service reads `SIGNED_IN`, the application's next OBO mint returns `consent-required`, and the only remedy was to sign in again in a different product's UI.

The sign-in interrupt is the honest place to ask — that prompt exists precisely because an application has said it needs to act on the user's behalf — so the dialog now offers an explicit, declinable checkbox and submits the answer with the sign-in.

**Offered checked, never silent.** Clicking "Log in" to continue a conversation is not by itself agreement that an application may act for you while you are away, so the grant is always visible and one click from being declined. Nothing is defaulted server-side: the DTO leaves it `undefined` and the mapper forwards exactly what the caller sent, including an explicit `false`. Reviewers may reasonably want the default flipped — it is a one-line change and I have no strong claim on it.

The OAuth choice is made before the redirect and the code exchange happens after it, so the value rides in the popup's redirect state (`ToolsetRedirectState.offlineUsageConsent`) — the callback route has no other way to learn it. The API-key path submits it directly.

Toolsets are deliberately untouched here: `offlineUsageConsent` is a field of Core's shared `ResourceSignInRequest` and `/v1/ops/toolset/signin` has the same gap. Filed as #8505 with a follow-up PR stacked on this one, since they share the same consent control.

**Verification:** `chat-api` 168 files / 2784 tests pass, `SigninInterruptDialog` spec passes, eslint and prettier clean, and `tsc --noEmit` output is byte-identical to `development` for both apps. Not exercised against a live DIAL deployment.

Issues:

- Issue #8504

**UI changes**

One checkbox added to the sign-in interrupt dialog, below the service rows: **"Allow use while I'm away"**, with the hint *"Lets this application use the connection on your behalf when you are not here — required for scheduled or background runs. You can revoke it by signing out."*

**Checklist:**

- [x] the pull request name starts with `fix(<scope>):`, `feat(<scope>):`, `feature(<scope>):`, `chore(<scope>):`, `hotfix(<scope>):` or `e2e(<scope>):`
- [x] the pull request name ends with `(Issue #<TICKET_ID>)`

### DIFF
--- a/apps/chat-api/src/external-services/dto/external-service.dto.ts
+++ b/apps/chat-api/src/external-services/dto/external-service.dto.ts
@@ -1,5 +1,12 @@
 import { ApiProperty, ApiPropertyOptional } from '@nestjs/swagger';
-import { IsEnum, IsNotEmpty, IsString, ValidateIf } from 'class-validator';
+import {
+  IsBoolean,
+  IsEnum,
+  IsNotEmpty,
+  IsOptional,
+  IsString,
+  ValidateIf,
+} from 'class-validator';
 
 export enum ExternalServiceAuthType {
   None = 'NONE',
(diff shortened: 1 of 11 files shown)


PolinaGurinovich97:
Closing as superseded by #8507.

#8507 ("feat(chat): request offline usage consent when signing in to a toolset (Issue #8505)") was stacked on this branch and merged into `development` on 2026-09-01 as `3e87d6a5c`, carrying every change from this PR along with the toolset side.

Verified by diffing this PR's head against `development`: the diff contains only deletions — no additions. All 11 files here are already on `development`, byte-identical (`external-service.dto.ts`, `external-services.mapper.ts` and its spec, `useExternalServiceLogin.ts`, `server-api/external-services.ts`, `translation-keys.ts`, `en.json`, `oauth/models.ts`). In `ToolsetAuthCallback.tsx` and `oauth/popup.ts`, `development` is ahead of this branch: it forwards `offlineUsageConsent` on the toolset path too, which this PR does not.

The `code_checks` failure here (`useExternalServiceLogin.spec.ts:153`, the extra sixth argument to `navigateToolsetOAuthPopup`) was fixed in that same merge — the spec on `development` now expects the trailing `undefined`.

Nothing left to rebase or merge.

PR #8534, opened 2026-09-01
feat(chat)!: redesign the conversation transfer toast around files and real progress

PolinaGurinovich97:
**Description:**

Redesigns the conversation import/export toast to the new Figma frame, and moves it into `@epam/ai-dial-chat-shared` so a second application can reuse it.

Each row now names the **file** being written or read — file-type icon plus the name, truncated with a tooltip — instead of the conversation title, and carries its own **determinate** ring rather than a shared aggregate bar that only counted settled jobs. Cancelling keeps the row on screen as `Canceled`. A failed row states the reason on the alert icon in place of the old bare retry button.

Behaviour worth calling out:

- **Progress is a monotonic weighted percentage**, not raw work units. The attachment count is unknown until the conversation has been fetched, so raw units would jump from 100% back to 9% the moment ten attachments were discovered. Fixed per-kind phase weights (prepare / transfer / finalize) let a discovered count subdivide the work still to do instead of moving the indicator backwards. `setJobProgress` clamps, refuses a lower value, and ignores writes to an already-settled job, so an aborted run unwinding in the background cannot advance a cancelled row.
- **Cancel is distinct from dismiss.** `cancelJob` aborts the work and leaves the row with status `Canceled`; `dismissJob` still removes it. A `Canceled` job needs no close confirmation — the user already chose to stop it — but does suppress the 8-second auto-close, so the outcome is always seen.
- **`FileTooLarge` fails eagerly.** The limit is on summed *input* bytes but guards *peak heap*: the archive pipeline holds roughly three copies at once (every attachment, `zipSync`'s output, and the copy handed to `Blob`), so the 512 MiB default peaks near 1.5 GiB. Sizing it against the ~2 GiB single-`ArrayBuffer` ceiling would have been the wrong bound. Exposed as `maxArchiveBytes` so a host can tune it without a library release.
- **The queue moved to `chat-shared`**, beside the `ConversationTransferJob` contract it renders, so a host can show transfer progress without taking on `conversation-panel`'s `react-window` and `@epam/ai-dial-sidebar` dependencies. `ConversationTransferErrorCode` moved with it — `conversation-panel` could not otherwise reach it.

Accessibility: the cancel control is hover-revealed but always mounted and focusable, so it is reachable by keyboard; the failure reason is the alert icon's `aria-label`, not only its tooltip, because a tooltip renders nothing on a mobile screen; the ring exposes `aria-valuenow`/`aria-valuetext` and is deliberately **not** mirrored in RTL, since a counter-clockwise sweep reads as work being undone.

Specs follow the OpenSpec lifecycle under `openspec/changes/redesign-transfer-queue-toast/` — one new capability (`conversation-transfer-progress`), one renamed (`conversation-panel-transfer-queue-ui` → `transfer-queue-ui`, all requirements removed and restated under `chat-shared` ownership), and deltas for `chat-hooks-conversation-transfer`, `conversation-export` and `conversation-import`.

Issues:

- Issue #<TICKET_ID>

**UI changes**

https://www.figma.com/design/t46nfThjIus5aUpMt43stA/DIAL-Chat-2.0--Chat?node-id=300-4752

Not yet checked in a real browser — see the reviewer notes below.

**BREAKING CHANGES**

| Before | After |
| --- | --- |
| `ImportExportQueue` from `@epam/ai-dial-conversation-panel` | from `@epam/ai-dial-chat-shared` (no compatibility re-export, per `remove-cross-package-reexports`) |
| `onDismiss` / `onRetry` | `onCancel` |
| `labels.allConversationsJobLabel`, `closeJobAriaLabel`, `retryJobAriaLabel` | `cancelJobAriaLabel`, `canceledLabel`, `jobErrorMessage`, `jobProgressAriaLabel`, `jobProgressValueText` |
| `ConversationTransferJob` | gains required `fileName` and `progress`; optional `errorCode` |
| `ConversationTransferJobStatus` | gains `Canceled` |
| `--cp-transfer-queue-*` | `--ieq-*`; the ring uses `--cprog-*` |

`retryJob` stays on `useConversationExport` / `useConversationImport` for hosts that want to re-expose it.

Known downstream consumer: `pg-chat` imports `ImportExportQueue` from `@epam/ai-dial-conversation-panel` and will need the new package versions plus the import-path and CSS-variable rename.

**Reviewer notes — not done in this PR**

- No manual browser pass: reaching the queue needs a running DIAL Core and an authenticated session. The four row states, hover-to-cancel, `mobile`/`desktop`, and `dir="rtl"` still want a human look. Keyboard reachability of cancel *is* asserted in jsdom (three `Tab` presses then `Enter`).
- `npm run verify:changed` is green — typecheck, lint and tests across every affected project. `npm run verify:full` additionally fails on `@epam/chat-api` and `mcp-app-sandbox` with 796 typecheck errors that reproduce identically on a clean tree, so they pre-date this branch.
- `npm run validate:docs` reports only the pre-existing `useGridEditingScroll` drift in `libs/chat-hooks/README.md`.

**Checklist:**

- [ ] the pull request name ends with `(Issue #<TICKET_ID>)` (comma-separated list of issues)


libs/conversation-panel/src/components/ImportExportQueue/ImportExportQueue.module.scss
```diff
@@ -1,36 +1,52 @@
...
 .textSecondary {
   color: var(
-    --cp-transfer-queue-text-secondary,
+    --ieq-text-secondary,
     var(--text-secondary, #57647a)
   );
 }
 
+.divider {
+  border-block-end: 1px solid
```
PolinaGurinovich97, line 17: review

libs/chat-shared/src/models/conversation-transfer.ts
```diff
@@ -22,9 +23,64 @@ export type ConversationTransferSubject =
     }
   | { kind: ConversationTransferSubjectKind.All };
 
+/** Reason a transfer job failed. Translation-free; hosts map each member to their own copy. */
+export enum ConversationTransferErrorCode {
```
PolinaGurinovich97, line 29: types

libs/chat-api-client/src/generated/README.md
```diff
@@ -3,23 +3,27 @@
...
+- Node.js
+- Webpack
+- Browserify
 
 Language level
-* ES5 - you must have a Promises/A+ library installed
-* ES6
+
+- ES5 - you must have a Promises/A+ library installed
+- ES6
```
valerydluski, line 14: check prettier

valerydluski (approved):

PR #8592, opened 2026-09-03
feat(conversation-panel): expose a corner-radius slot for the row button

PolinaGurinovich97:
## What

Adds `ConversationPanelStyles.itemRadiusClassName` — the corner-radius class applied to the button inside each conversation row, defaulting to `'rounded-xl'` (12px).

## Why

The radius was hardcoded on the row's button, so a host whose design calls for a different one had to reach into the rendered DOM from its own stylesheet:

```scss
/* what a consumer had to write */
li[class*='group/conversation'] > a > button {
  border-radius: 8px;
}
```

That selector encodes this library's internal element structure and its `group/conversation` marker class — it breaks the moment either changes. `styles.searchWrapperClassName` and `styles.itemIconBadgeClassName` already cover the same need elsewhere in the panel; this closes the gap for the row itself.

## How

- `panel-props.ts` — the new `itemRadiusClassName` slot on `ConversationPanelStyles`.
- Threaded exactly like `itemIconBadgeClassName`: `ConversationPanel` (row props + dep array) → `RowStyles` in `virtual-row.ts` → `RowRenderer` → `ConversationRow`.
- `ConversationRow` — `itemRadiusClassName = 'rounded-xl'`, applied in place of the literal.

Two deliberate boundaries:

- **Only the radius is themable.** The row's layout utilities (`h-8 w-full justify-start gap-2 py-2 ps-0`) stay hardcoded and are always applied, so a caller cannot silently lose the row's geometry.
- **The slot replaces its default rather than merging**, matching `searchWrapperClassName` and `itemIconBadgeClassName`. Two competing `rounded-*` utilities on one element would resolve by stylesheet order, not by the order they appear in the class attribute.

The name carries the radius on purpose: a general `itemClassName` would read like an open class hook, which makes both boundaries above surprising.

Not a breaking change: an optional slot with the previous value as its default, so a caller that sets nothing renders exactly as before.

## Testing

- `vitest run --root libs/conversation-panel` — 138 tests / 8 files pass.
- `ConversationRow.spec.tsx` covers the default, and the override case asserts the default class is *gone* (not merely appended to) while `h-8` / `w-full` survive.
- `ConversationPanel.spec.tsx` asserts `styles.itemRadiusClassName` reaches the row through the whole chain, which is where a wiring mistake would otherwise hide.
- `eslint`, `prettier --check`, and `tsc -b` on the lib and spec projects are clean. `npm run validate:docs` reports only the pre-existing `libs/chat-hooks/README.md` / `useGridEditingScroll` failure already present on `development`.
- README updated in the same change, alongside the `searchWrapperClassName` paragraph.


PolinaGurinovich97:
Superseded by #8593, which themes the row, the search field and the New chat button through CSS custom properties (`--cp-row-radius`, `--cp-search-radius`, and the ui-kit's `--radius-control`) instead of adding a class slot per control.

PR #8604, opened 2026-09-03
fix: close nine QA divergence findings and reconcile four stale spec clauses

PolinaGurinovich97:
Triage of the 22-finding QA divergence report against the current source. The findings split four ways: four were already fixed on `development`, six turned out to be spec drift rather than defects, nine are real defects fixed here, and three are parked because the code that owns them lives in a third-party package.

## Code fixes

| Finding | Fix |
| --- | --- |
| TC-UI-0045 | 404 page: the reduced-motion block pinned the gradient headline's `background-position` but never stopped its animation, which overwrote the pin every frame. `.code` now joins `animation: none`. |
| TC-SKILL-0022 | Skill editor: the mobile Files accordion defaulted to expanded, pushing the Instructions editor below the fold on the narrowest layout. |
| TC-CAT-0048 | Toolset auth: the currently-selected auth-type option stayed enabled while logged in, so a no-op control looked actionable. All three options disable together now, and each exposes `aria-pressed`. |
| TC-TASK-0016 | Scheduled tasks sort: the active option carried no programmatic state. Options are now `menuitemcheckbox` rows with `aria-checked` plus a check icon; the menu closes explicitly so picking a sort still dismisses it. The trigger falls back to `labels.sortLabel` instead of rendering a nameless button when the sort key matches no option. |
| TC-TASK-0040 | Builder form: `isSubmitDisabled` covers both "not ready" and "already submitting", so it could not distinguish them. A new `isSubmitting` prop adds a spinner, `aria-busy`, and a live-region announcement, with the button's accessible name held stable. Wired through the scheduled-task form and the prompt editor. |
| TC-ATT-0052 | Attachment tile: the uploading overlay's `aria-label` sat on a role-less `div` and never reached the accessibility tree. It is an indeterminate `role="progressbar"` now, the tile reports `aria-busy`, and `uploadingLabel` is threaded from the app instead of resting on its English default. |
| TC-CONV-0367 | Confirmation dialogs: closing one dropped focus to `<body>`. Restoration lives in `useAsyncConfirmDialog`, so revoke, unshare, unpublish, rename and delete are all covered; the conversation row passes its kebab trigger as the target, because the clicked menu item unmounts with its menu. |
| TC-PUB-0018 | Publish panel: a one-frame focus guard was standing in for a focus trap, so Tab walked out of the open modal. Tab and Shift+Tab wrap inside the panel now, and are left alone while focus sits in one of its portals. |
| TC-UI-0237 | Announcement banner: title and description shrank in proportion, leaving the description a sliver of ellipsis instead of collapsing out of view. |

## Spec corrections

The code was right in these four; the spec had gone stale and the failing assertions were written against it.

- **`markdown-code-blocks`** — the copy button keeps a stable accessible name and announces through a live region (changed deliberately in c868c870c); the spec still demanded an `aria-label` swap. Covers TC-CHAT-0074 and TC-CHAT-0079.
- **`catalog-item-details-fetch`** — unlimited usage rows deliberately render no progress bar, so a panel whose rows are all unlimited legitimately mounts zero `role="progressbar"` nodes. Covers TC-CAT-0054 and TC-CAT-0051.
- **`catalog-item-details-fetch`** — the `Summary` component and the `item.summary` / `item.intro` fields were removed in the 1.0 redesign (5688649db), but a requirement still described them, contradicting two other clauses in the same file. Covers TC-CAT-0022.
- **`file-manager-folder-creation`** — the folder-name rules are evaluated in a fixed order owned by `@epam/ai-dial-react-file-manager`, and a multi-fault name reports the earliest matching rule, so `../secret` reports consecutive dots rather than a forbidden character. Covers TC-FILE-0099.

## Not addressed here

Three findings reproduce but their defect is outside this repository, so they stay open:

- **TC-FILE-0125** — the batch-upload gate is in `@epam/ai-dial-react-file-manager`: on any conflict `onUploadFiles` is never called and the whole batch waits. `useDialFileUploadBatch` already derives `uploadMode` per file and uploads immediately — it simply is not invoked.
- **TC-TASK-0045** — `MarkdownEditor` from `@epam/ai-dial-ui-kit` accepts no `aria-label`, `textareaProps` or `id`, so its `textarea` has no accessible name. `PromptEditor` has the same defect by construction.
- **TC-FILE-0275** — `treeOptions.header` feeds `CollapsibleSidebar.title`, which the kit renders only as vertical text while collapsed. An expanded tree panel has no header by design; this needs a product decision rather than a patch.

## Also

Restores the `useGridEditingScroll` re-export from `chat-hooks`, which `AGENTS.md` and the lib README both promise. It had been dropped when the hook moved to `chat-shared`, and `npm run validate:docs` was failing on it before this branch.

## Test plan

- Unit tests added for all nine fixes, in `attachment-input`, `builder-form` (via `scheduled-tasks`), `chat-hooks`, `publish-panel`, `scheduled-tasks`, `skill-editor` and `apps/chat`.
- `lint`, `typecheck` and `test` green across every affected project and `@epam/chat`.
- `npm run validate:docs` passes.
- READMEs updated for `attachment-input`, `builder-form`, `chat-hooks`, `conversation-input`, `publish-panel` and `scheduled-tasks`.

## Note for QA

Six scenarios change their expected assertion and will keep failing if re-run unchanged: TC-TASK-0016 (`menuitemcheckbox` + `aria-checked`, not `aria-selected`), TC-TASK-0040 (`aria-busy` with the button's name unchanged), TC-ATT-0052 (a `progressbar` now exists), TC-CAT-0048 (assertion widens), TC-CONV-0367 (focus returns to the kebab trigger, not the menu item), plus all four spec cases.


libs/scheduled-tasks/src/components/ScheduledTasks/ScheduledTasks.tsx
```diff
@@ -98,8 +105,47 @@ export const ScheduledTasks: FC<ScheduledTasksProps> = ({
...
+  const sortItems = labels.sortOptions.map((option) => {
+    const isActiveSort = option.key === sortKey;
+
+    return {
+      ...option,
+      selectable: true,
+      checked: isActiveSort,
+      renderItem: () => (
+        <>
+          <span className="flex-1 truncate text-start">{option.label}</span>
```
hhharmEpm: Maybe use EllipsisTooltip?

hhharmEpm (approved):

PR #8623, opened 2026-09-04
feat: Add inline\full width mode for MCP App, do a workaround with calling tool directly

hhharmEpm:
**Description:**

- biggest change - if MCP endpoint is available, call the tool directly, not use tool result from conversation. It makes the initial MCP App open slower (request is duplicated) but it's the only workaround until Core starts to send the real tool call result
- added full\inline mode support
- added caching of result while conversation is open

Issues:

- Issue #<TICKET_ID>

**UI changes**

inline
image
full 
image

**Checklist:**

- [X] the pull request name complies with [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/)
- [X] the pull request name starts with `fix(<scope>):`, `feat(<scope>):`, `feature(<scope>):`, `chore(<scope>):`, `hotfix(<scope>):` or `e2e(<scope>):`. If contains breaking changes then the pull request name must start with `fix(<scope>)!:`, `feat(<scope>)!:`, `feature(<scope>)!:`, `chore(<scope>)!:`, `hotfix(<scope>)!:` or `e2e(<scope>)!:` where `<scope>` is name of affected project: `chat`, `chat-e2e`, `overlay`, `shared`, `sandbox-overlay`, etc.
- [X] the pull request name ends with `(Issue #<TICKET_ID>)` (comma-separated list of issues)
- [X] I confirm that do not share any confidential information like API keys or any other secrets and private URLs


libs/mcp-apps/src/hooks/useMcpAppInlinePreview/useMcpAppInlinePreview.ts
```diff
@@ -0,0 +1,151 @@
...
+ * Builds `hostContext` with `displayMode: 'inline'` so the app can render a
+ * more compact layout than it would inside the full-width canvas.
+ *
+ * `cache` (shared with `useOpenMcpAppCanvas` via the same
+ * `useMcpAppResponseCache` instance, keyed by `cacheKey`) is checked before
+ * fetching — switching from this preview to the full canvas (or back) for
+ * the same message reuses the same fetch/live-tool-re-call instead of
+ * repeating it. Call the returned `reload` to bypass and refresh it.
+ */
+export const useMcpAppInlinePreview = (
```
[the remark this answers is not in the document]
hhharmEpm: it's tmp

apps/chat/src/hooks/attachment/useMcpAppInlinePreview.ts
```diff
@@ -0,0 +1,168 @@
...
+          );
+        }
+        if (cancelled) return;
+        setHtml(fetchedHtml);
+        setToolResult(resolvedResult);
+        setStatus(McpAppInlinePreviewStatus.Ready);
+      } catch {
+        if (!cancelled) setStatus(McpAppInlinePreviewStatus.Error);
+      }
+    })();
```
valerydluski: Could we avoid the async IIFE inside `useEffect` and extract this logic into a `useCallback` instead? It would make the effect cleaner and easier to follow.

libs/mcp-apps/src/hooks/useMcpAppInlinePreview/useMcpAppInlinePreview.ts
```diff
@@ -0,0 +1,168 @@
...
```
valerydluski, line 44: Could we move this hook into `chat-hooks`? It looks generic enough to live in our hooks library instead of keeping the implementation here.

apps/chat/src/components/ConversationView/McpAppInlinePreview/McpAppInlinePreview.tsx
```diff
@@ -0,0 +1,126 @@
...
+ * sized to the mounted app's actual content height. A header strip above the
+ * app — outside its rendered content, so it never overlaps whatever the app
+ * draws — carries a reload button (re-fetches from scratch, bypassing
+ * `cache`) and the expand-to-canvas button (`onExpand`,
+ * `useOpenMcpAppCanvas`, which reuses the same cache entry this preview
+ * populated). Renders nothing while the MCP Apps sandbox proxy isn't
+ * configured, matching `useOpenMcpAppCanvas`'s own no-op behavior in that
+ * case.
+ */
+const McpAppInlinePreview: FC<Props> = ({
```
PolinaGurinovich97: can we movve to lib?

libs/mcp-apps/src/hooks/useMcpAppInlinePreview/useMcpAppInlinePreview.ts
hhharmEpm: it's in mcp-apps library now, so can be reused. we can move to chat-hooks, if someone else will need this logic, though it seems to be mcp-app specific

PolinaGurinovich97 (approved):

PR #8748, opened 2026-09-11
fix: replace ReDoS-prone trailing trims and update the nx toolchain to 23

PolinaGurinovich97:
## Description of changes

Two independent pieces of work, cleanly separated by commit.

### 1. ReDoS hardening (1 commit, 19 files)

Replace polynomial-ReDoS trailing-trim regexes with linear scans in `chat-hooks` (`string-utils`, `file-name`, `skill`, `useSkillEditorLoad`, `dial-file-manager-path.util`) and `conversation-stages` (`stage-name`), with tests covering the trim behaviour.

### 2. Toolchain update: nx 22.7.5 → 23.2.1 (5 commits)

- 27 nx migrations applied, and all 6 deferred prompt migrations worked through. Most were already satisfied because the repo was ahead of them — already on Vite 8, Vitest 4, React 19, flat ESLint config and TypeScript 6.
- Removed a duplicate `@nx/vitest` plugin entry in `nx.json`: the migration moved `testTargetName` off the deprecated `@nx/vite/plugin` without noticing an `@nx/vitest` entry already provided it. The surviving entry is the one carrying `ciTargetName`/`testMode`, which is what atomizes `test-ci`.
- Two `overrides` added for peer-resolution conflicts. `@nx/react@23` newly declares an optional `express@^4` peer while this repo runs express 5 — there is no module federation here, so nx's express code paths never load. Separately, `attachment-canvas`'s caret range drifted onto a newly published `0.2.0-rc` line of `@epam/ai-dial-react-pdf-highlighter`, whose ui-kit peer range cannot admit a 0.14 prerelease (npm only matches a prerelease against a comparator carrying a prerelease at the same `major.minor.patch`). That one is also pinned in the lib's own manifest, because `overrides` do not apply to consumers of the published package.
- `prettier` 3.8 → 3.9 reformatting; 3.9 collapses short union types onto one line. All lint errors were this and auto-fixable — no rule-level changes were needed.
- `--node-env` → `--config-node-env` for both webpack apps and the `start:api:dev` script, since webpack-cli 7 renamed the flag. Verified behaviour-preserving rather than merely silenced: webpack-cli assigns `process.env.NODE_ENV` before loading the config, so `configuration.mode` is still derived from it.
- Raised the build heap ceiling for `chat-hooks`, whose 17 API Extractor passes (`dts({ rollupTypes: true })`) now peak between 4 and 5 GB, just past the default old-space limit.
- Version tables updated in `README.md` and `docs/architecture.md`.

## Verification

| Check | Result |
| --- | --- |
| `nx run-many -t build --all` | 34/34 pass |
| `nx run-many -t test --all` | 31/31 projects pass (`chat` 2106 tests, `chat-api` 3026 tests) |
| `nx run-many -t lint --all` | 34/34 pass, 0 errors |
| `npm run validate:docs` | pass |
| `nx run-many -t typecheck --all` | fails — pre-existing, see below |

`typecheck` reports 310 errors, confined to `chat-api` (309) and `mcp-app-sandbox` (1); the other 32 projects are clean. These are **pre-existing and not caused by this branch** — the error set is identical with and without the migration's tsconfig changes. They accumulated unnoticed because CI does not run `typecheck` at all. Most are `chat-api` spec files not seeing `describe`/`it`/`expect`, because `apps/chat-api/tsconfig.spec.json` sets `types: ["node", "vitest"]` where Vitest needs `vitest/globals` for ambient globals.

## Known follow-ups (not addressed here)

- The `add-ignore-deprecations-for-ts6` migration pinned `esModuleInterop: false`, `types: ["*"]` and `noUncheckedSideEffectImports: false` into `tsconfig.base.json`. That migration assumes a workspace arriving from TypeScript 5, whereas this repo was already on TS 6, so the pins reintroduce pre-TS6 defaults it had moved past. They provably cause none of the current failures, but `esModuleInterop: false` is a latent trap.
- The `chat-hooks` heap bump raises a ceiling rather than fixing the cause. Dropping `rollupTypes` would remove the 17 API Extractor passes and would match the `preserveModules` JS output, but it changes the published type layout, so it belongs in its own change with the package-boundary tests and consumer fixture exercised.
- `npm audit` on this branch reports one high finding: `brace-expansion` resolves to `1.1.16` from the lockfile, while the two current DoS advisories require `>= 1.1.18`. `npm update brace-expansion` lifts it without touching the 2.x and 5.x copies in the tree. (`smol-toml`, which arrived transitively with nx 23, is already covered by the `smol-toml: 1.8.0` override.)
- CI (`pr.yml`) runs neither `typecheck` nor `lint`, and builds only four libs. Both breakages found while verifying this update — the webpack-cli flag and the `chat-hooks` OOM — sit in exactly that blind spot.

## Applicable issues

Neither the ReDoS hardening nor the nx toolchain update has a tracked issue.

## UI changes

None — security hardening and tooling only.

## Checklist


ai-dial-actions:
Hey there and thank you for opening this pull request! 👋🏼

We require pull request titles to follow the [Conventional Commits specification](https://www.conventionalcommits.org/en/v1.0.0/) and it looks like your proposed title needs to be adjusted.

Details:

```
No release type found in pull request title "Fix/polynomial redos string trims". Add a prefix to indicate what kind of release this pull request corresponds to. For reference, see https://www.conventionalcommits.org/

Available types:
 - feat
 - fix
 - docs
 - test
 - ci
 - chore
```

valerydluski (approved):

PR #8767, opened 2026-09-14
feat(conversation-panel): add hiddenSources prop to fully exclude a source

Alexander-Kezik:
## Description of changes

Lets a host drop conversations by source (e.g. Organization) before tab filtering, search, and grouping, so a hidden source disappears from every tab (including All) and its pill in FilterTabs — not just its own tab, unlike the existing isFilterTabsHidden. Also widens the FilterTabs pill gap from 4px to 8px.

## Applicable issues

- fixes #<ISSUE_ID>

## UI changes

## Checklist

<details>

1. type (required)
   - `feat` - A new feature
   - `fix` - A bug fix
   - `chat`
   - `overlay`
   - `shared`
   - `sandbox-overlay`
   - `visualizer-connector`

</details>


libs/conversation-panel/src/components/FilterTabs/FilterTabs.tsx
```diff
@@ -29,26 +31,34 @@ const TABS: { value: FilterTab; labelKey: FilterTabLabelKey }[] = [
...
-          selected={activeTab === value}
-          onClick={() => onChange(value)}
-          className={mergeClasses(
-            /* The filter row splits its width evenly and keeps the pill silhouette. */
-            'flex-1 justify-center rounded-full',
-            tabClassName,
-          )}
-        />
-      ))}
+      {TABS.filter(({ value }) => !hiddenSources?.includes(value)).map(
```
hhharmEpm: maybe move filtering to the variable and useMemo?
```
const tabs= useMemo(() => TABS.filter(({ value }) => !hiddenSources?.includes(value)), [hiddenSources])
```

Alexander-Kezik: Extracted into a named tabs variable, but skipped useMemo: TABS is a fixed 4-entry array so filtering it is trivial, and the filtered result is only ever iterated straight into JSX (tabs.map(...)) — it's never passed down to a child component as a prop, so there's no risk of unnecessary child re-renders/effects from a new array reference. Also, the intended caller (ConversationPanelView in pg-chat) constructs hiddenSources fresh on every render anyway, so a useMemo keyed on it wouldn't actually cache anything — it'd just add bookkeeping without avoiding work.

hhharmEpm (approved):

Issue #8707, opened 2026-09-10
Expose public classes for the Chat components to provide the customisation capabilities.


Kuznietsov:
### Description

# Request to Dial: stable public classNames for host styling

**Packages:** `@epam/ai-dial-*`, `@epam/ai-dial-ui-kit`  
**Repos:** [epam/ai-dial-chat](https://github.com/epam/ai-dial-chat) (`development`), [epam/ai-dial-ui-kit](https://github.com/epam/ai-dial-ui-kit)

## Problem

Composable Dial UI is embedded in a host app that needs custom chrome (panel, composer, attachment tiles, model menu). Dial CSS modules hash class names on build (e.g. `_userBubble_cq829_1`, `_modelSelectorButton_1py46_24`), so host overrides use fragile selectors:

- `[class*='…']` substrings of CSS-module locals
- DOM order / `:nth-child` / `:has(textarea)`
- `role` + default `aria-label` strings
- Tailwind utilities that are **not** present in the published ui-kit CSS (`size-[84px]`, `p-1.5`, `end-1`, `top-1`, `group-hover/attachment-tile`)

These break on Dial package upgrades (already observed for ConversationInput action-row order and AttachmentCard sizing).

## Ask

Please expose **stable public classNames** (and/or documented style props) on the elements below so hosts can target them without relying on hashed module names or DOM structure.

Suggested naming convention: `dial-<pkg-short>-<element>` (e.g. `dial-ci-action-row`, `dial-cm-user-bubble`, `dial-ai-attachment-tile`).

Optional: also ship the missing Tailwind utilities in `@epam/ai-dial-ui-kit`, or move AttachmentCard dimensions into CSS modules so hosts do not need utility polyfills.

---

### 1. Conversation panel / sidebar

| Fragile selector | Why it is needed | Dial source | Suggested public className |
| --- | --- | --- | --- |
| `[role='complementary']` | Remove end border so host panel owns outline | [`SidebarPanel.tsx` L185–186](https://github.com/epam/ai-dial-chat/blob/development/libs/sidebar/src/components/SidebarPanel/SidebarPanel.tsx#L185-L186) (`@epam/ai-dial-sidebar`) | `dial-sb-aside` |
| `[role='complementary'] > div:first-child` | Header padding | [`Header.tsx`](https://github.com/epam/ai-dial-chat/blob/development/libs/sidebar/src/components/Header/Header.tsx) (first child of aside) | `dial-sb-header` |
[...]

PR #8830, opened 2026-09-15
feat(libs): stable public dial-* class names on every library (Issue #8707)

PolinaGurinovich97:
## Why

A host embedding the Composable UI has nothing stable to style against. CSS-module locals are hashed at build time, Tailwind utilities move whenever the layout changes, and DOM order is not a contract — so hosts end up on selectors that break on every package upgrade:

- `[class*='_userBubble_']`, `[class*='_modelSelectorButton_']` — substrings of module locals
- `> div:nth-child(2) > button`, `:has(textarea)` — DOM order
- `[role='complementary']`, `[role='list'][aria-label='Attached files']` — accessibility contracts, and in the `aria-label` case a plain English string that breaks the moment the app is translated

Reported in [#8707](https://github.com/epam/ai-dial-chat/issues/8707) with an inventory of the selectors a host had resorted to.

## What

Every library that renders UI now stamps stable, host-addressable class names on the elements a host actually styles. **25 libraries, 66 classes.** Each library owns one `src/constants/public-class-names.ts` record, re-exported from its entry point, so a host imports the names instead of hardcoding them:

```tsx
import { CONVERSATION_INPUT_CLASS } from '@epam/ai-dial-conversation-input';

CONVERSATION_INPUT_CLASS.actionRow; // 'dial-ci-action-row'
```

```css
.dial-ci-action-row {
  flex-wrap: nowrap;
}
```

The classes **carry no declarations of their own** — no lib stylesheet selects on them, verified against the built CSS. Adding one is therefore never a visual change, and a host's rule needs no `!important` it would not otherwise need.

### Naming

`dial-<lib-prefix>-<element>[-<state>]`, flat kebab-case, state additive rather than replacing. The prefix is the **library's directory name** (`dial-catalog-card`, `dial-scheduled-tasks-card-grid`, `dial-settings-panel-tab`).

The first draft of this change derived the prefix from each library's CSS-custom-property prefix, so the class contract and the theming contract would read alike. That does not generalise past the five libraries it was written for, for two measured reasons:

- **Five var prefixes are claimed by two libraries each** — `--ai-` (`attachment-input`, `catalog`), `--cm-` (`conversation-messages`, `chat-shared`), `--cc-` (`quotations`, `attachment-canvas`), `--pp-` (`prompts`, `publish-panel`), `--sp-` (`settings-panel`, `source-panel`).
- **Eleven of the twenty-two libraries with stylesheets use more than one own prefix** — `publish-panel` alone uses seven.

The five short prefixes #8707 proposed by name (`dial-sb-`, `dial-cp-`, `dial-cm-`, `dial-ci-`, `dial-ai-`) are kept and documented as a **closed list**; nothing new joins it.

### Deliberate departures from the issue

- **`dial-cm-code-block` is `dial-chat-shared-code-block`.** `dial-cm-` belongs to `conversation-messages`, while `MarkdownCodeBlock` is rendered from `chat-shared`. One prefix owned by two libraries is the collision the grammar exists to prevent. The `conversation-messages` README points at the new name.
- **No `styles.classNames` prop API.** The issue offered it as an alternative; the classes are emitted unconditionally instead, which the issue also offered and which needs no call-site change. `styles` already means theming tokens in these libraries, and portalled or deeply nested elements are unreachable by a prop.
- **`useModelSelector` never accepted `selectedItemClassName` / `selectedItemCheckClassName`.** The issue asked for those to be "surfaced"; only `searchHeaderClassName` existed. There was nothing to surface.
- **The selected model row's check mark has no class here** — `@epam/ai-dial-ui-kit` draws it. It gained `dial-kit-menuitem-check` in the kit; see below.

### Scope choices worth review

- **`catalog` gets five classes, not one per component.** With 44 components and 24 stylesheets, a class per component is a contract too large to keep accurate. The set is the layout skeleton — root, toolbar, both card kinds, list view — and card internals are reached by descending from the card's class, which the README states.
- **The virtualised grid box in `CardGrid` has no class.** Its height is recomputed every scroll frame, so a host styling it fights the virtualizer rather than the design.
- **Where the drawn box moves, the class follows it.** A `MenuItem` with a `rightControl` draws the row rectangle on the wrapper, so the class sits there; `ProgressBar` and `ListView` return their track and their empty state as roots of their own, and both paths are stamped and asserted.

## Not in this PR

The issue's second half — the Tailwind utilities a published package does not ship (`size-[84px]`, `p-1.5`, `end-1`, `group-hover/attachment-tile`) — **is not solved here.** An earlier revision published the design-token theme as `@epam/ai-dial-chat-shared/tailwind-preset` so a host could satisfy the `content`-glob requirement; that was reverted on request, and the theme stays in the repo-root `tailwind.config.js` as its single owner. A host still has to run Tailwind CSS 3 and scan each package's `dist`, but the token theme this repo builds against is not published, so semantic utilities (`bg-layer-raised`, `text-secondary`) remain unavailable to it.

## Guard tests

A lost public class fails the way the styling guide's dead-style checks describe: the build passes, types pass, lint passes, and a host's stylesheet silently stops applying. Every class therefore has at least one test that **locates the element by role, label or text first and only then asserts the class** — finding it *by* the class would pass with the class on the wrong node.

That discipline paid for itself repeatedly. Cases the guard tests caught, not review:

- `conversation-stages` — the class first landed on `StagesPanel`'s row toggle, which is the collapse control of a retry-attempt row rather than the group a host collapses.
- `attachment-canvas` and `source-panel` — both were documented as sitting on the same element as `dial-sb-aside`. `SidebarPanel` puts `styles.className` on the **wrapper**; the `aside` inside it carries `dial-sb-aside`. The source-panel test had passed only because its own `SidebarPanel` stub put the class on the `aside`, a shape the real panel never renders.
- Three stubs (`IconButton` in `navigation-panel`, `EditorSection` in `toolset-editor`, `SidebarPanel` in `source-panel`) silently dropped or relocated `className`, which turns a class assertion into a no-op.

## Adjacent defects fixed

Found while verifying, each with its own commit message:

- **`McpAppInlinePreview` rendered with no background.** Its card used `bg-layer-2`, a design-system 1.0 token the 2.0 theme does not define, so Tailwind emitted no rule at all.
- **`@epam/ai-dial-mcp-apps` shipped a broken `./styles.css`.** Without `lib.cssFileName`, Vite named the CSS asset after `lib.name`, so the CSS-module rules landed in `ai-dial-mcp-apps.css` while the exported `index.css` held only the appended utility layer. The lib is now themable through `McpAppInlinePreviewColors` and its stylesheet actually reaches the exported file.
- **The breakpoint-sync guard read a literal from a file that no longer held it**, failing `code_checks` on this branch.
- **`IconPlus` in `NewChatButton` had no `aria-hidden`**, which `.claude/rules/a11y.md` requires of a decorative icon inside an already-named control.

## Companion PR

The kit half is [epam/ai-dial-ui-kit#882](https://github.com/epam/ai-dial-ui-kit/pull/882): public class names on every design-system 2.0 component, including the four `dial-kit-*` names #8707 asked for and `dial-kit-menuitem-check` for the chosen-row glyph. It is independent — this PR does not depend on that release.

## Verification

- `typecheck`, `lint`, `format` — clean
- `build:affected` — passing; note it typechecks spec files through `vite-plugin-dts` while the `typecheck` target does not, which is what surfaced three spec-only type errors
- `validate:docs` — passing; all 25 library READMEs carry the contract section
- A repo-wide audit over all 25 records: every one of the 66 classes is stamped, every one is asserted by at least one test, none appears as a string literal in a component file, and none carries a declaration in any lib stylesheet

## Follow-ups (not blocking)

- Extend `scripts/validate-docs.mjs` with the audit above, so the contract is CI-enforced rather than review-enforced.
- The composer's desktop `flex-nowrap` action row — the issue's "ideal fix" — needs design sign-off.
- `openspec/config.yaml` has three `rules` entries whose unquoted `": "` makes YAML parse them as maps, so OpenSpec silently drops every `proposal` and `tasks` rule for every change in this repo. One-line fix, unrelated to this work.


libs/chat-shared/tailwind-preset.cjs
```diff
@@ -0,0 +1,232 @@
...
+ *
+ * The `.cjs` extension is required, not stylistic: this package is ESM
+ * (`"type": "module"`), and Tailwind configs are CommonJS loaded by tooling
+ * outside the bundler. It carries no imports, so `chat-shared` stays a
+ * `type:shared` library that imports nothing.
+ */
+
+// Default color palette is light when no themes presented
+
+const backgroundsColors = {
```
PolinaGurinovich97: ?

libs/mcp-apps/src/components/McpAppInlinePreview/McpAppInlinePreview.tsx
```diff
@@ -100,13 +102,23 @@ const McpAppInlinePreviewBase: FC<McpAppInlinePreviewProps> = ({
   }
 
   return (
-    <div className="bg-layer-2 flex w-full min-w-0 flex-col overflow-hidden rounded-xl border border-tertiary">
+    <div
+      className={mergeClasses(
+        MCP_APPS_CLASS.preview,
+        'bg-layer-2 flex w-full min-w-0 flex-col overflow-hidden rounded-xl border border-tertiary',
```
PolinaGurinovich97: scss variable

